PR #15, opened 2020-04-22
Pull request for master

tonellotto:
## syntax sugar

### DIFF
--- a/pyterrier/__init__.py
+++ b/pyterrier/__init__.py
@@ -20,7 +20,7 @@ def init(version=None, mem=None, packages=[], jvm_opts=[], redirect_io=True, log
         version(str): Which version of Terrier to download. Default=None.
             If None, find the newest Terrier version in maven and download it.
         mem(str): Maximum memory allocated for java heap in MB. Default is 1/4 of physical memory.
-        packages(list(str)): Extra maven package coordinates files to load. Default=[]. More information at http://terrier.org/docs/v5.0/terrier_develop.html
+        packages(list(str)): Extra maven package coordinates files to load. Default=[]. More information at https://github.com/terrier-org/terrier-core/blob/5.x/doc/terrier_develop.md
         jvm_opts(list(str)): Extra options to pass to the JVM. Default=[].
         redirect_io(boolean): If True, the Java System.out and System.err will be redirected to Pythons sys.out and sys.err. Default=True.
         logging(str): the logging level to use.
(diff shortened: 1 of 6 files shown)


pyterrier/__init__.py
```diff
@@ -15,17 +11,22 @@
...
     Function necessary to be called before Terrier classes and methods can be used.
     Loads the Terrier.jar file and imports classes. Also finds the correct version of Terrier to download if no version is specified.
 
     Args:
         version(str): Which version of Terrier to download. Default=None.
             If None, find the newest Terrier version in maven and download it.
-        mem: Maximum memory allocated for java heap in MB. Default=4096.
-        packages: Extra .jar files to load.
+        mem(str): Maximum memory allocated for java heap in MB. Default=4096.
+        packages(list(str)): Extra .jar files to load. Default=[].
```
cmacdonald, line 23: this should refer to Maven package coordinates, not jar files.

pyterrier/__init__.py
```diff
@@ -15,17 +11,22 @@
...
     """
     Function necessary to be called before Terrier classes and methods can be used.
     Loads the Terrier.jar file and imports classes. Also finds the correct version of Terrier to download if no version is specified.
 
     Args:
         version(str): Which version of Terrier to download. Default=None.
             If None, find the newest Terrier version in maven and download it.
-        mem: Maximum memory allocated for java heap in MB. Default=4096.
-        packages: Extra .jar files to load.
+        mem(str): Maximum memory allocated for java heap in MB. Default=4096.
```
cmacdonald, line 22: Please, lets leave None as default. I think JVM will use 25% of system RAM by default, which is less brittle than a hard-coded number.

tests/test_trec_indexer.py
```diff
@@ -26,6 +26,7 @@ def tearDown(self):
 
     def test_TREC_indexing(self):
         indexer = pt.TRECCollectionIndexer(self.test_dir)
+        pt.redirect_stdouterr()
```
cmacdonald, line 29: Was this intended? It should be default now

pyterrier/__init__.py
tonellotto: Ok. I just left the message as written.

pyterrier/__init__.py
tonellotto: Ok, but we should clearly state that in the doc: 4 GB is not the default on my server, according to the 25% rule.

pyterrier/__init__.py
cmacdonald: While we are here, we should improve things!

tests/test_trec_indexer.py
tonellotto: yep. debugging leftover.

cmacdonald:
Thanks. Once tests pass, I will merge.

Issue #8, opened 2020-04-12
indexers targeting memory indices


cmacdonald:

cmacdonald:
Addressed; pt.IndexingType.MEMORY now works.

PR #17, opened 2020-04-22
Added singlepass TREC indexers (both basic and block)

tonellotto:

### DIFF
--- a/pyterrier/bootstrap.py
+++ b/pyterrier/bootstrap.py
@@ -1,4 +1,3 @@
-import os
 import deprecation
 
 from . import mavenresolver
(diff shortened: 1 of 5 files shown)


cmacdonald:
A few comments:
1. Does your patch work for the DFIndexer too? Can you test it there as well?
2. Can you assert that a direct index is and isnt created in the correct places
3. Your PR seems to include all of the commits in #15. Can you rebase? A new PR might be needed.
4. I wish to add support for creating memory indices - see #8. i wonder if the constructor parameter should be called type, and support the following values:
type="classic" (default)
type="singlepass"
type="memory"
Should we have an [enum](https://www.tutorialspoint.com/enum-in-python) to make it easier for users?

tonellotto:
Ok. I'll close and request a new pull once I address the comments.

PR #22, opened 2020-05-02
Fixed error in index

tonellotto:
When trying to use CollectionFactory to set the collection class(es) to use in indexing, I get an error while trying to use the "_class" attributes. The proposed changes make the indexing tests pass on my machine (MacOS, Python 3.7).


cmacdonald:
As discussed, I think this is addressed in jnius 1.3.0, which has now been released.

PR #26, opened 2020-05-20
Fixed BM25_pr

ArthurCamara:
BM25_pr example is not working. Second argument is a dict, not a string. Fixed to make it explicit.


cmacdonald:
Good spot, thanks!

PR #32, opened 2020-06-08
Fix name of qrels for TREC-Covid

albertoueda:
It was not allowing get_residuals(qrels) to work properly for round2


cmacdonald:
good spot, thanks!

Issue #7, opened 2020-04-12
DfIndexer should set properties for metadata indexing


cmacdonald:

cmacdonald:
Resolved.

PR #35, opened 2020-06-21
IterDictIndexer

seanmacavaney:
Motivation: The existing `DFIndexer` requires entire collection to be loaded into memory, which could be a problem for large collections. It also doesn't currently support indexing multiple fields or customizing which fields are stored in the meta index [1]. This PR addresses both issues by allowing the user to send an iter of dict objects, where each key in the dict corresponds to a field to be indexed or a meta field (controlled by `field` and `meta` parameters). I simply use the `org.terrier.indexing.FlatJSONDocument` to handle all the field stuff, rather than rolling my own implementation because it seems to handle the cases I foresee. Tests included.

[1] Aside: Is meta index field functionality broken for `DFIndexer`? The code seems like it was [meant to support this](https://github.com/terrier-org/pyterrier/blob/master/pyterrier/index.py#L227), but the tests [show that it doesn't work](https://github.com/terrier-org/pyterrier/blob/master/tests/test_dfindex.py#L46).

### DIFF
--- a/examples/notebooks/indexing.ipynb
+++ b/examples/notebooks/indexing.ipynb
@@ -580,6 +580,71 @@
     {
       "cell_type": "markdown",
       "metadata": {
+        "id": "zjSUuBcou_L2",
+        "colab_type": "text"
+      },
+      "source": [
+        "## Indexing a iterable, generator, etc.\n",
+        "\n",
+        "You may not want to load all documents into memory, particularly for large collections. Terrier can index iterable objects (e.g., generators) that yield `dict` objects.\n",
+        "\n",
+        "To do thise, we can use a `pt.IterDictIndexer()` object. By default, `text` will be indexed and `docno` will be stored in the meta index. These can be configured with the `fields` and `meta` parameters, respectively."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "CBYEKxK4u_uV",
+        "colab_type": "code",
+        "colab": {
+          "base_uri": "https://localhost:8080/",
+          "height": 1000
+        },
+        "outputId": "f1880b9e-495e-44f2-a094-dbbf5ae7d1d9"
+      },
+      "source": [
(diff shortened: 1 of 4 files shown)


seanmacavaney:
Ah, I see my aside above is related to #7.

cmacdonald:
Excellent idea, it always troubled me that DFIndexer required everything in memory.

I shall review and come back to you if I have questions. Thank you. 

I wonder if this can replace the DFIndexer? or can DFIndexer call this for every row? (the only change is df.iterrows() vs iter)

seanmacavaney:
Yeah, it could probably be generalized to accept dicts, named tuples, or pd.Series. Then one could also pass in `indexer.index(df.itertuples())` or `indexer.index(df.iterrows())`. Could also just detect if a dataframe is passed in and to that for ya.

cmacdonald:
I have addressed #7 last night. Would you able to add an examplar usage in https://github.com/terrier-org/pyterrier/blob/master/examples/notebooks/indexing.ipynb
(you can update the notebook in Colab and then save and commit to your branch) then I will merge. I'm OK with both indexers existing for now.

Dont worry about Github Action tests not passing on your branch - that has been resolved on the master branch now fixed.

seanmacavaney:
Should be set now

cmacdonald:
Thanks, looks good. 

About metadata sizes - we compress the metadata for each document using zlib, but clearly seeks can take longer if things are excessive; I will add an parameter to allow these to be specified in the constructor.

seanmacavaney:
gotcha, thanks!

cmacdonald:
@seanmacavaney I would propose to move `fields=` handling to the Pyterrier indexer constructor. Any objections?

seanmacavaney:
Sounds reasonable - while we're at it, might as well also include the metadata fields? It would also be nice to have an easy way to provide the reverse meta keys. `meta_forward=` and `meta_reverse=`?

Issue #37, opened 2020-06-25
evaluate using pytrec_eval is slow


cmacdonald:

cmacdonald:
Mostly is about converting qrels and results to dictionaries; I have sped up pt.Experiment() by reusing the qrels dictionary. We cannot improve until https://github.com/cvangysel/pytrec_eval/issues/22 is fixed.

seanmacavaney:
#48 should help with this in the meantime.

cmacdonald:
Closing, as #48 and #50 address most of the issues. Thanks @seanmacavaney

PR #48, opened 2020-07-11
speeding up convert_qrels_to_dict

seanmacavaney:
Related to #37. Speeding up convert_qrels_to_dict 100x using itertuples instead of iterrows. itertuples is faster because it avoids boxing the rows in pd.Series. I'm surprised it made so much of a difference here. Maybe due to the large number of rows in my test file (311,410).

My timing code:

```
import pyterrier; pyterrier.init()
df = pyterrier.Utils.parse_qrels('robust04.qrels')

import timeit
result = timeit.timeit(lambda: pyterrier.Utils.convert_qrels_to_dict(df), number=5)
print(result) # total time for 5 executions in seconds
```

Before (with iterrows)
```
142.19985203596298
```

After (with itertuples)
```
1.3646108560496941
```

### DIFF
--- a/pyterrier/utils.py
+++ b/pyterrier/utils.py
@@ -213,9 +213,9 @@ class Utils:
         Returns:
             dict: {qid:{docno:label,},}
         """
-        run_dict_pytrec_eval = defaultdict(dict)     
-        for index, row in df.iterrows():
-            run_dict_pytrec_eval[row['qid']][row['docno']] = int(row['label'])
+        run_dict_pytrec_eval = defaultdict(dict)
+        for row in df.itertuples():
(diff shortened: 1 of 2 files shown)


cmacdonald:
Great observation. I hadn't appreciated the difference in speed. I will review shortly. This improvement could help other areas of the platform.

cmacdonald:
I note from https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.itertuples.html, that _"On python versions < 3.7 regular tuples are returned for DataFrames with a large number of columns (>254)."_
I don't think that should be a problem. 

We can do similar improvements for runs.

PR #51, opened 2020-07-13
Keep only qid, query, docno on intersect, Fix #49

### LINKED ISSUE #49: Preserve "query" column when using set union/intersection operators
this code works:
```python
(br1 & br2).transform(topics)
```

but this one does not:
```python
(br1 & br2) >> br3
br3.transform(topics)
```

albertoueda:
No repeated columns anymore. However, fields such as scores could be interesting to final user to decide how to combine them, avoiding the necessity of additional steps/retrievals. I'm not sure if this is intended, so I followed the previous idea of remove repeated columns.

### DIFF
--- a/pyterrier/transformer.py
+++ b/pyterrier/transformer.py
@@ -244,9 +244,9 @@ class SetIntersectionTransformer(BinaryTransformerBase):
     
     def transform(self, topics):
         res1 = self.left.transform(topics)
-        res2 = self.right.transform(topics)
-        # NB: there may be other duplicate columns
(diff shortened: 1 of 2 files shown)


cmacdonald:
Hi Alberto, 

Can you provide an updated unit test. Also, there is an existing unit test that is now failing?

Also the queries might be different (e.g. query expansion from one side). Perhaps we keep the query from the left hand side?

Craig

cmacdonald:
This looks good. I think however we need to clearly document the semantics of each operator, in terms of columns and rows, in the doc-strings. Would you able to do to that Alberto, and I can pick that up and merge?

cmacdonald:
LGTM, thanks Alberto

PR #53, opened 2020-07-17
addresses move methods in Utils into better name #5

cmacdonald:
Creates a new modules called io, which contains:
 - read_topics
 - read_qrels
 - read_results & write_results

These methods variously support various formats, for instance:
 - read_topics supports "trec", "trecxml", "singleline" 
 - read_results & write_results support "trec" and "letor"

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ See examples in the [indexing notebook](examples/notebooks/indexing.ipynb) [![Op
 # Retrieval and Evaluation
 
 ```python
-topics = pt.Utils.parse_trec_topics_file(topicsFile)
-qrels = pt.Utils.parse_qrels(qrelsFile)
+topics = pt.io.read_topics(topicsFile)
+qrels = pt.io.read_qrels(qrelsFile)
 BM25_br = pt.BatchRetrieve(index, controls={"wmodel": "BM25"})
 res = BM25_br.transform(topics)
 pt.Utils.evaluate(res, qrels, metrics = ['map'])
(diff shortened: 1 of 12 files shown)


cmacdonald:
@albertoueda could you review this PR?

albertoueda:
> @albertoueda could you review this PR?

LGTM. Some comments about my practical usage:

1. Besides get_qrels, get_topics, I've added a `get_extra_info(variant, sep, names)` which returns a simple pandas df with some generic [sep]-delimited file, with the columns as names if desired. It is useful for files other than topics and qrels and with no specified format, or even topics files which do not follow a pre-defined format.
2. The `read_topics` and `read_qrels` could have a optional parameter of a parsing function, giving to the user the possibility of write his own file_to_dataframe function.
3. The `read_results` and `write_results` could also be used to easily reproduce experiments with baselines (e.g., read runs 1,2,3 and show the results considering 1 as baseline) and allow users to apply previous results to other pipelines (similarly as cache), such as read a old run and put it into a re-rank pipeline.

cmacdonald:
> 1.
I cant see particular scenarios that would be useful.

> 2. The read_topics and read_qrels could have a optional parameter of a parsing function, giving to the user the possibility of write his own file_to_dataframe function.

I think we have a format key. I could use the format key to lookup a dictionary. If users want to add more formats, they simply add an entry to the dictionary. I cant see much utility in calling read_topics if we simply call the user-supplied function.  I will update the branch on this basis.

> 3. The read_results and write_results could also be used to easily reproduce experiments with baselines.

I think this is a separate issue - can we use a dataframe as a source for a pipeline, or as a run in Experiment. Can you file a new issue.

cmacdonald:
Revised code using a dictionary for mapping format keys to parsing methods, addresses point 2 of Alberto feedback

PR #57, opened 2020-07-30
Added TREC PrecisionMedicine and Covid datasets

albertoueda:


albertoueda:
The failed test `test_br` passes in my local repository. Should I suggest a PR to the master (updated) instead?

cmacdonald:
I have partially merged this patch. 
Notes:
 - I didnt merge dataset.get_extra_info(), as this feels dataset specific, and you can use dataset.get_one_file() and parse manually.
 - "Allow users to specify corpus.home" -- I cant see HOW this allow users to specify the home. Unless pt.get_dataset() gets a kwargs `corpus_home` as well.
If you feel strongly, we can cover the latter under a separate issue.

Issue #90, opened 2021-02-12
get_dataset() -> ir_datasets


cmacdonald:

cmacdonald:
Addressed by #93, thanks Sean!

PR #93, opened 2021-02-15
ir_datasets integration #90

seanmacavaney:
Added `IRDSDataset` and populated it with everything available in ir_datasets (with `irds:` prefix).

I used the "variant" mechanism to select which fields to use, particularly for topics. (E.g., some topics have title, description, narrative fields, `get_topics(variant=)` selects which one.)

I added a `get_corpus_iter()` function that returns an iterator over document dicts (verbose=False disables a default tqdm bar). This can be used by IterDictIndexer:

```
>>> import pyterrier as pt
>>> pt.init()
>>> dataset = pt.datasets.get_dataset('irds:cord19')
>>> indexer = pt.index.IterDictIndexer('./cord19')
>>> indexer.index(dataset.get_corpus_iter(), fields=('title', 'abstract'))
[INFO] [starting] https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/historical_releases/cord-19_2020-07-16.tar.gz
[INFO] [finished] https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/historical_releases/cord-19_2020-07-16.tar.gz: [03:26] [3.66GB] [17.8MB/s]
[INFO] [starting] building docstore                                                                                                      
[INFO] [starting] extracting tarfile
[INFO] [finished] extracting tarfile [6.71s]
docs_iter: 192509it [00:11, 16264.41it/s]
[INFO] [finished] docs_iter: [00:11] [192509it] [16264.34it/s]
[INFO] [finished] building docstore [11.84s]
cord19 documents: 100%|████████████████████████████████████████████| 192509/192509 [01:06<00:00, 2899.25it/s]
14:56:39.186 [main] WARN  o.t.structures.indexing.Indexer - Indexed 60 empty documents
<org.terrier.querying.IndexRef at 0x7f3bde33a230 jclass=org/terrier/querying/IndexRef jself=<LocalRef obj=0x55b9e1313e50 at 0x7f3bde2b2270>>
>>> idx = pt.IndexFactory.of(pt.IndexRef.of(indexer.path))
>>> idx.lexicon['covid'].frequency
200582
```

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -214,6 +214,109 @@ class RemoteDataset(Dataset):
         return "RemoteDataset for %s, with %s" % (self.name, str(list(self.locations.keys())))
 
 
+class IRDSDataset(Dataset):
+    def __init__(self, irds_id):
+        self._irds_id = irds_id
+        self._irds_ref = None
+
+    def irds_ref(self):
+        if self._irds_ref is None:
+            self._irds_ref = ir_datasets.load(self._irds_id)
+        return self._irds_ref
+
+    def get_corpus(self):
+        raise NotImplementedError("IRDSDataset doesn't support get_corpus; use get_corpus_iter instead")
+
+    def get_corpus_iter(self):
+        ds = self.irds_ref()
+        assert ds.has_docs(), f"{self._irds_id} doesn't support get_corpus_iter"
+        for doc in ds.docs_iter():
+            doc = doc._asdict()
+            # pyterrier uses "docno"
+            doc['docno'] = doc.pop('doc_id')
+            yield doc
+
+    def get_index(self, variant=None):
+        # TODO: should this automaticaly build an index?
+        raise NotImplementedError("IRDSDataset doesn't support get_index")
+
+    def get_topics(self, variant=None):
+        """
+            Returns the topics, as a dataframe, ready for retrieval. 
+        """
+        ds = self.irds_ref()
+        assert ds.has_queries(), f"{self._irds_id} doesn't support get_topics"
+        qcls = ds.queries_cls()
+        assert variant is None or variant in qcls._fields[1:], f"{self._irds_id} only supports the following topic variants {qcls._fields[1:]}"
+        df = pd.DataFrame(ds.queries_iter())
+
+        df.rename(columns={"query_id": "qid"}, inplace=True) # pyterrier uses "qid"
+
+        if variant is not None:
+            df.rename(columns={variant: "query"}, inplace=True) # user specified which version of the query they want
+            df.drop(df.columns.difference(['qid','query']), 1, inplace=True)
+        elif len(qcls._fields) == 2:
+            # auto-rename single query field to "query" if there's only query_id and that field
+            df.rename(columns={qcls._fields[1]: "query"}, inplace=True)
+        else:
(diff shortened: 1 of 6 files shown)


cmacdonald:
Looks good. Could we update docs/datasets.rst as well?

seanmacavaney:
Sure- updated docs/datasets.rst. Thanks for catching the typo as well

cmacdonald:
Good work. A few comments for discussion - minor quibbles.

One disadvantage is that we now have more datasets than can be listed by Pandas in its default. Should we suppress max row display for this dataframe only?

datasets

Should we add some more attributes to make them more easily browsable:
 - link to the description URL for each dataset
 - a column depicting language. we might filter by English by default?

 - Should we add an indexing example using `get_corpus_iter()`? 
 - I think I should probably add an `get_corpus_iter()` to all Pyterrier datasets that have a corpus.

seanmacavaney:
That all sounds reasonable.

> Should we suppress max row display for this dataframe only

Is there a way to set display options for a single dataframe? I can only find the global `pd.set_option('display.max_rows', x)` option. Maybe just a higher global limit like 200? 500?

> link to the description URL for each dataset

- [ ] Seems reasonable

> a column depicting language. we might filter by English by default?

There are some challenges there when a dataset is not completely homogenous. I explore those challenges a bit [in this issue](https://github.com/allenai/ir_datasets/issues/19).

Seems right to filter down to only homogenous English datasets though, as there's not yet much consideration of other languages in pyterrier.

> Should we add an indexing example using get_corpus_iter()?

- [ ] Yup, I'll add

cmacdonald:
I thought max rows was also accessible through df.style, but now I'm not so sure. Lets see how many datasets we have once the filtering is applied.

seanmacavaney:
Filtering to only English corpora and topics gets us from 158 down to 128.

(Note: this change requires a new version of ir_datasets so you'll need to `pip install --upgrade ir_datasets` or use the requriements.txt file.)

seanmacavaney:
okay, this is ready for another review (and re-consideration of setting max pandas rows).

For:

> I think I should probably add an get_corpus_iter() to all Pyterrier datasets that have a corpus

it should just be a matter of constructing a pyterrier document iterator? I could take a stab at this if you like.

cmacdonald:
OK, I added some changes for the PyTerrier side of things to give MSMARCO and Antique corpus iters. I need to make a corpus_iter for vaswani (and other TREC corpora); we can file that as a separate issue.

The generated documentation lists the full table.
I wonder if we can link to the full table, and/or display a warning if some rows are suppressed in pt.list_datasets()

```python
if pd.get_option("pd.maxrows") < len(df)
  warning.warn("Pandas is display less rows. Perform pd.BLA to increase the number of rows, or refer to the full table at https://pyterrier.readthedocs.io/en/latest/datasets.html#available-datasets")
```

pyterrier/datasets.py
```diff
@@ -718,6 +743,9 @@ def _merge_years(self, component, variant):
     "info_url" : "https://trec.nist.gov/data/precmed.html",
 }
 
+def trec2text(dataset):
```
seanmacavaney, line 746: delete this for now and make a new issue for it?

cmacdonald: yes, woops

cmacdonald:
Remaining failing test is unrelated to this PR. Merging.

PR #100, opened 2021-02-23
support for IRDSDatasets in pt.text.get_text()

seanmacavaney:


cmacdonald:
Looks pretty good. Do we need to document lookup times for docnos and docids?

seanmacavaney:
Something like this?

Here's lookup times on `irds:cord19/trec-covid` for title+abstract fields. For fair comparisons with retrieve w/ metadata, it's the entire pipeline process, including retrieval. Times reported are the medians of 5 runs. Cold start clears the FS cache between runs, warm start does not. Index and IRDS Docstore both on HDD.

batch_retrieve w/ metadata is always the fastest, but get_data(dataset) gets close when the filesystem cache helps out.

It's unclear why `get_data(index)` is all that much slower than `batch_retrieve w/ metadata`. It's doing essentially the same thing, no? I guess it also avoids making another copy of the dataframe, but that shouldn't be _that_ expensive.

|                                    | Cold Start | Warm Start |
|------------------------------------|-----------:|-----------:|
| batch retrieve                     |      7.44s |      6.63s |
| batch retrieve w/ metadata         |      7.78s |      7.11s |
| batch retrieve >> get_data(index)   |     10.20s |      9.68s |
| batch retrieve >> get_data(dataset) |      8.45s |      7.22s |

cmacdonald:
This is much more detailed than what i imagined!
reverse docno lookups are expensive - possibly binary searching the meta index. I think your impl assumes that there are enough docnos to keep in memory.

seanmacavaney:
But `get_data(index)` uses the docids returned from batch retrieve, though, no? https://github.com/terrier-org/pyterrier/blob/master/pyterrier/text.py#L65

IRDS does a binary search over a mmap'd file.

PR #130, opened 2021-03-31
pipeline parallelisation, including pickling of IndexRef, BatchRetrieve objects etc

cmacdonald:
Example use case - retrieval parallelisation with joblib:

```python
dph = pt.BatchRetrieve(index)
br_par = dph.parallel(4, backend='joblib')
pt.Experiment(
	[dph, br_par],
	topics2019,
	qrels2019,
	["num_q", "num_ret", "map", "mrt"])
```

Example use case - retrieval parallelisation with [ray](https://ray.io):

```python
import ray
ray.init(address='auto')
dph = pt.BatchRetrieve(index)
br_par = dph.parallel(4, backend='ray')
pt.Experiment(
	[dph, br_par],
	topics2019,
	qrels2019,
	["num_q", "num_ret", "map", "mrt"])
```

Theoretically, ray allows parallelisation of retrieval across many machines. I havent been able to successfully test that though (when ray, some things segfault later).


cmacdonald:
And I see that it segfaults here too.

cmacdonald:
Segmentation faults fixed here.

docs/parallel.rst
```diff
@@ -0,0 +1,80 @@
...
+rather than just BatchRetrieve, as then only the queries and final results  need to be passed betwene processes. Indeed among the
+following semantically equivalent pipelines, we expect `parallel_pipe0`  and `parallel_pipe2`  to be faster than `parallel_pipe1`::
+
+    parallel_pipe0 = pt.BatchRetrieve(index, metadata=["docno", "text"]).parallel() >> pt.text.sliding() >> pt.text.scorer() >> pt.text.max_passage()
+    parallel_pipe1 = ( pt.BatchRetrieve(index, metadata=["docno", "text"]).parallel() >> pt.text.sliding() ).parallel(2)  >> pt.text.max_passage()
+    parallel_pipe2 = pipe.parallel(2)
+
+
+There are of course overheads on paralllelisation - for instance, the Terrier index has to be loaded for *each* separate process, 
+so your machine(s) require enough memory.
```
seanmacavaney, line 70: Maybe also worth mentioning neural transformers here specifically too? Usually, there the limiting factor is the GPU (either in terms of memory or compute), so parallelizing there wouldn't have any gains, or at worst would crash due to out-of-memory errors.

Unless the machine has multiple GPUs, in which parallel() could also maybe set `CUDA_VISIBLE_DEVICES` accordingly on each sub-process. I'm not familiar with joblib or ray, so I don't know how easy that is to do.

seanmacavaney (approved):
This is pretty dope. lmk what you think about the note in the documentation about the impact on neural transformers.

PR #132, opened 2021-04-06
linear combination of queries

cmacdonald:


pyterrier/rewrite.py
```diff
@@ -309,3 +309,65 @@ def transform(self, queries_and_docs):
...
+
+    def __init__(self, weights : List[float], format : str = 'terrierql', **kwargs):
+        super().__init__(**kwargs)
+        self.weights = weights
+        self.format = format
+        if format not in ["terrierql", "matchopql"]:
+            raise ValueError("Format must be one of 'terrierql', 'matchopql'")
+
+    def _terrierql(self, row):
+        return "(%s)^%f (%s)^%f" % (
```
seanmacavaney, line 349: should we be concerned about escaping the query text, is it safe to assume it's already in the target query language?

cmacdonald: I think its safe to assume its already in the correct query language.

seanmacavaney: alright, then lgtm

seanmacavaney (approved):

cmacdonald:
Thanks Sean

PR #149, opened 2021-04-24
using ir_measures aggregators

seanmacavaney:


cmacdonald:
Thx, I had been going to ask if irm could do the aggregation

seanmacavaney:
There's a big caveat now that it doesn't do the `-c` behavior--- it only averages over the intersection of the queries in the run and qrels. Is this a deal-breaker for merging into master?

cmacdonald:
I'm not sure what PyTerrier does. It will I guess interact with the "drop_unused" option, which drops topics not in the qrels;  IIRC, the problem is for settings where the system doesnt return any results qrels (regardless of whats in the topics).

seanmacavaney:
This can happen inadvertently though. E.g., for conjunctive queries where there's no intersection between the posting lists.

I guess with most benchmarks these days, this won't be a problem though. We can fix this edge case later.

PR #159, opened 2021-05-12
Update pandas version in requirements.txt

albertoueda:
There is a bug in Pandas < 0.25.0 that adds the string "None" as a suffix to column names on merging. That impacts pipelines such as those using `CombSumTransformer`, causing an error on searching columns such as "score" (because it became "scoreNone" instead).

Issue describing the error: 
https://github.com/pandas-dev/pandas/issues/24782

```python
>>> result = df.merge(df, left_index=True, right_index=True, suffixes=(None,'_dup'))
>>> result
   0None  0_dup
0      1      1
>>> result.columns
Index(['0None', '0_dup'], dtype='object')
```

There is also the possibility of other recent versions (current is 1.2.4), I'm using the 1.0.1 and it looks fine for now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy
-pandas
+pandas>=0.25.0
 wget
 pytrec_eval>=0.5
 tqdm


cmacdonald:
Thanks for the report Alberto. Could we have a unit test for the CombSumTransformer that fails on 0.24?

albertoueda:
> Thanks for the report Alberto. Could we have a unit test for the CombSumTransformer that fails on 0.24?

ok, will do so!

cmacdonald:
bump?

PR #172, opened 2021-05-27
dynamic allocation of IRDSDataset

seanmacavaney:
Background: [CLIRMatrix](https://www.aclweb.org/anthology/2020.emnlp-main.340/) is [being added to ir_datasets](https://github.com/ssun32/ir_datasets), which contains over 100k sub-datasets. There are so many due to the large number of language combinations. Rather than allocating all of them in the registry (which would be expensive to do every time the library is loaded), we're introducing a "pattern matching" approach that builds them on-the-fly, as they are requested. Thus, not all datasets will appear directly in the registry anymore, which means that they would not be available in pyterrier without a change like this one.

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -844,6 +844,12 @@ def get_dataset(name, **kwargs):
     """
         Get a dataset by name
     """
+    # Some datasets in ir_datasets are built on-the-fly (e.g., clirmatrix).
+    # Handle this by allocating it on demand here.
+    if name not in DATASET_MAP and name.startswith('irds:'):
+        # remove irds: prefix
+        ds_id = name[len('irds:'):]
+        DATASET_MAP[name] = IRDSDataset(ds_id)
     rtr = DATASET_MAP[name]
     rtr._configure(**kwargs)
     return rtr


seanmacavaney:
Just also realized this also supports the case where [an extension dataset](https://github.com/seanmacavaney/dummy-irds-ext) is loaded after pyterrier is initialized:

```python
import pyterrier as pt
pt.init()
import dummy_irds_ext
```

cmacdonald:
Seems an easy merge. I was thinking we need to have "Search for datasets" somehow in PyTerrier, as the `pt.list_datasets()` is now too long to show all. (and should that pipeline be implemented directly in PyTerrier)

seanmacavaney:
No need to force an IRDS version-- it just uses the existing `ir_datasets.load` function.

Would it be easy to add a dataset search feature to the documentation? I'm not super familiar with the capabilities of rest.

In code, I think this would be as easy as a new `find_datasets()` function, which would just basically do:

```python
def find_datasets(query):
    datasets = list_datasets()
    return datasets[datasets['dataset'].str.contains(query)]
```

Maybe we want to add tags to make the size of the list more manageable? A few that come to mind immediately are "popular", "downloadable", and "english".

PR #196, opened 2021-06-24
add + and len() support on a index.

9meo:
fix conflict and update _has_positions


cmacdonald (changes requested):
This looks excellent Sarawoot. One detail: if len(index) returns 0, the test would still pass. Can we ensure len(index) == index.getCollectionStatistics().getNumberOfDocuments()

cmacdonald:
Thanks for the contribution Sarawoot.

PR #200, opened 2021-07-03
From dataset136

cmacdonald:
Not finished, but could be reviewed.


pyterrier/batchretrieve.py
```diff
@@ -62,11 +63,70 @@ def __init__(self, verbose=0, **kwargs):
...
+    import json
+    
+    if isinstance(dataset, str):
+        dataset = get_dataset(dataset)
+    if version != "latest":
+        raise ValueError("index versioning not yet supported")
+    indexref = dataset.get_index(variant)
+
+    classname = clz.__name__
+    # now look for, e.g., BatchRetrieve.args.json file, which will define the args for BatchRetrieve, e.g. stemming
```
seanmacavaney, line 84: Will naming this BatchRetrieve.args.json make it annoying if/when we change the name of the class later?

pyterrier/batchretrieve.py
```diff
@@ -62,11 +63,70 @@ def __init__(self, verbose=0, **kwargs):
...
+    indexdir = indexref #os.path.dirname(indexref.toString())
+    argsfile = os.path.join(indexdir, classname + ".args.json")
+    if os.path.exists(argsfile):
+        with autoopen(argsfile, "rt") as f:
+            args = json.load(f)
+            # anything specified in kwargs of this methods overrides the .args.json file
+            for k,v in kwargs.items():
```
seanmacavaney, line 91: `args.update(kwargs)`?

pyterrier/batchretrieve.py
```diff
@@ -62,11 +63,70 @@ def __init__(self, verbose=0, **kwargs):
...
+                args[k] = v
+            kwargs = args
+    return clz(indexref, **kwargs)   
+                
 class BatchRetrieve(BatchRetrieveBase):
     """
     Use this class for retrieval by Terrier
     """
 
+    #from_dataset = staticmethod(partial(_from_dataset, clz=BatchRetrieve))
```
seanmacavaney, line 101: cleanup (also, the `partial` import at the top)

pyterrier/datasets.py
```diff
@@ -201,17 +203,51 @@ def _get_one_file(self, component, variant=None):
...
-        for (local, URL) in file_list:
+
+        direxists = os.path.exists(localDir)
+        
+        location = self.locations[component]
+        if is_lambda(location) or isinstance(location, types.FunctionType):
+            # functions are expensive to call, normally another HTTP is needed.
+            # just assume we have everthing we need if we have the local directory already
+            if direxists:
+                return localDir
```
seanmacavaney, line 222: Should most of this function be wrapped in a try/finally to clean up the dir if anything goes wrong?

cmacdonald: I was wondering about touching an empty file showing that download completion has occurred, and checking if that exists.

pyterrier/batchretrieve.py
cmacdonald: I think we can just make an additional json file and run that for a while.

pyterrier/batchretrieve.py
```diff
@@ -98,8 +96,7 @@ class BatchRetrieve(BatchRetrieveBase):
     Use this class for retrieval by Terrier
     """
 
-    #from_dataset = staticmethod(partial(_from_dataset, clz=BatchRetrieve))
-    @staticmethod
+        @staticmethod
```
seanmacavaney, line 99: indentation looks off here? (is this tabs vs spaces?)

pyterrier/datasets.py
seanmacavaney: that sounds less ugly, yeah

pyterrier/batchretrieve.py
cmacdonald: yes, I spotted. and ran unit tests before pushing again. this should be fixed now.

seanmacavaney (approved):

seanmacavaney (approved):

PR #211, opened 2021-07-27
only compute ranks for score column

seanmacavaney:
A bit of background: When `rank` is called for a full dataframe, it computes the ranks for every column. This is a bit of a waste because there's no need to have the ranks of other columns, e.g., docno, as they are immediately discarded. What's worse is that it can throw errors when columns exist in the dataframe that are not sortable, such as tensors.

`rank` can also be called on a series, so that's exactly what I change it to do.


cmacdonald:
Good find!

PR #215, opened 2021-08-09
Terrier weighting models expressed in Python

cmacdonald:


cmacdonald:
(unit tests pass once terrier-python-helper is installed.)

pyterrier/batchretrieve.py
```diff
@@ -570,6 +620,14 @@ def __setstate__(self, d):
         self.properties.update(d["properties"])
         for key,value in d["properties"].items():
             self.appSetup.setProperty(key, str(value))
+        #TODO consider the context state?
+
+    @staticmethod 
+    def from_dataset(dataset : Union[str,Dataset],
```
seanmacavaney, line 626: looks like this belongs in a different PR?

seanmacavaney (commented):
This is really cool!

Have you tested the performance difference between a pure java wmodel vs one of these? I'd expect there to be some impact, as there's overhead in jnius when switching between python and java. If it's noticeable, it might be worth mentioning this in the documentation.

Besides that, looks good to me.

pyterrier/batchretrieve.py
cmacdonald: I started from a from_dataset branch, but I think this bit is already merged.

cmacdonald:
Efficiency timings on Vaswani for the record:
Baseline in 2.2120 seconds
Lambda in 81.0745 seconds

seanmacavaney (approved):

Issue #205, opened 2021-07-07
Query biased document summary in PyTerrier


adambaker:
I'm trying to extract text snippets for a search results page. I've found the documentation page for running terrier in a JSP, which describes [some configuration for snippet extraction](https://github.com/terrier-org/terrier-core/blob/5.x/doc/terrier_http.md#using-the-web-based-interface).

I'm still not sure how to use this in pyterrier. It seems like `BatchRetrieve`'s `properties` or `control` parameters might be necessary to pass the required configuration, but then it's not clear to me how to get the query biased summary from the result set. Is there a way to do this with PyTerrier?

Thanks!

cmacdonald:
Hi adam,

The query-biased summariser in Terrier is that great. What about just breaking the text of the retrieved documents into sentences, and passing them through another text scoring ranking pipeline?

Craig

cmacdonald:
Adam, there is an draft PR for PyTerrier query biased summarisation at https://github.com/terrier-org/pyterrier/pull/223. We're not planning to merge for this week's PyT release, but it might be of interest.

cmacdonald:
The corresponding PR has been merged - documentation at https://pyterrier.readthedocs.io/en/latest/text.html#query-biased-summarisation-snippets.

PR #223, opened 2021-09-07
query biased summarisation poc - addresses #205

cmacdonald:


pyterrier/text.py
```diff
@@ -247,6 +247,67 @@ def slidingWindow(sequence : list, winSize : int, step : int) -> list:
...
+        psgres = tsp(docres)
+        if len(psgres) == 0:
+            print('no passages found in %d documents for query %s' % (len(docres), docres.iloc[0].query))
+            docres = docres.copy()
+            docres[summary_attr]  = ""
+            return docres
+
+        psgres[["olddocno", "pid"]] = psgres.docno.str.split("%p", expand=True)
+
+        newdf = psgres.groupby(['qid', 'olddocno'])[text_attr].agg('...'.join).reset_index().rename(columns={text_attr : summary_attr, 'olddocno' : 'docno'})
```
seanmacavaney, line 306: '...' should be joinstr?

cmacdonald:
needs test cases.

pyterrier/text.py
```diff
@@ -247,6 +247,67 @@ def slidingWindow(sequence : list, winSize : int, step : int) -> list:
...
+            >> pt.text.scorer(body_attr="text", wmodel='Tf', background_index=index)
+        )
+        
+        # use psg_scorer for performing query-biased summarisation on docs retrieved by br 
+        retr_pipe = br >> pt.text.qbs(psg_scorer)
+
+    """
+    import pyterrier as pt
+    tsp = (
+        pt.apply.query(lambda row: row['query'].replace('"', ""))
```
seanmacavaney, line 283: What's this for? If it's to remove query language, something more general is probably needed. And I'm not really sure why you'd want to

pyterrier/text.py
```diff
@@ -247,6 +247,67 @@ def slidingWindow(sequence : list, winSize : int, step : int) -> list:
...
+     - joinstr(str): how to join passages for a given document together
+
+    Example::
+
+        # retrieve documents with text
+        br = pt.BatchRetrieve(index, metadata=['docno', 'text'])
+
+        # use Tf as a passage scorer on sliding window passages 
+        psg_scorer = ( 
+            pt.text.sliding(text_attr='text', length=15, prepend_attr=None)
```
seanmacavaney, line 273: The quality of the summaries will depend a lot on this. In particular, with overlapping segments, text could appear twice. Really long segments could cause problems. You would never want to use it with prepend_attr (which is on by default). Etc. Maybe it's worth building pt.text.sliding into this transformer to enforce some of these?

seanmacavaney: I guess overlapping passages is important to guarantee phrases are captured. It would get a little messy, but a greedy strategy that makes sure that no passages with adjacent passage numbers were selected could resolve that problem.

cmacdonald: there are other alternatives such as using a passage index, such as using using a sentence splitter rather than sliding window. I think this is a PoC, anything else would be a different pipeline that is user changable

PR #228, opened 2021-09-16
pt.Experiment consistency in missing qids in topics/qrels/runs

### LINKED ISSUE #226: pt.Experiment consistency in missing qids in topics/qrels/runs
https://github.com/terrierteam/ir_measures/pull/17 changes the behaviour of ir_measures in the presence of qids that dont appear in the run but do appear in the qrels.

Some work is needed on pt.Experiment (and friends) to align with this change.

The general principle is that we should average in a safe manner.

There are different situations:

1. A qid is present in the topics and qrels, but not in the run - this topic should be given 0 score and therefore decrease the averaging. https://github.com/terrierteam/ir_measures/pull/17 does this for us.
2. A qid is present in the qrels, but not in the topics. 
  - ir_measures WILL count these qids during evaluation, as it does NOT have knowledge of the topics.
  - pt.Experiment should should ignore such topics during retrieval (by dropping them from the qrels), and ensure that they do not contribute to evaluation scores. 
3. A qid is present in the topics, but not in the qrels. Currently, pt.Experiment executes the topics, but does not count in evaluation. The `drop_unused` can be used to drop such topics. This is the status quo, and is unchanged.

This issue concerns the implementation of (2) above.

seanmacavaney:
fixes #226

### DIFF
--- a/docs/experiments.rst
+++ b/docs/experiments.rst
@@ -160,25 +160,25 @@ The ``filter_by_qrels`` and ``fitler_by_topics`` parameters control the behaviou
 do not perfectly overlap. When ``filter_by_qrels=True``, topics are filtered down to only the ones that have qids in the
 qrels. Similarly, when ``fitler_by_topics=True``, qrels are filtered down to only the ones that have qids in the topics.
 
-For example, consier qrels that include qids ``A`` and ``B`` and topics that include ``B`` and ``C``. The results with
+For example, consier topics that include qids ``A`` and ``B`` and qrels that include ``B`` and ``C``. The results with
 each combination of settings are:
 
 +----------------------+----------------------+------------------+--------------------------------------------------------------------+
 | ``filter_by_topics`` | ``filter_by_qrels``  | Results consider | Notes                                                              |
 +======================+======================+==================+====================================================================+
-| ``True`` (default)   | ``False`` (default)  | ``B,C``          | ``A`` is removed because it does not appear in the topics.         |
+| ``True`` (default)   | ``False`` (default)  | ``A,B``          | ``C`` is removed because it does not appear in the topics.         |
 +----------------------+----------------------+------------------+--------------------------------------------------------------------+
 | ``True`` (default)   | ``True``             | ``B``            | Acts as an intersection of the qids found in the qrels and topics. |
 +----------------------+----------------------+------------------+--------------------------------------------------------------------+
 | ``False``            | ``False`` (default)  | ``A,B,C``        | Acts as a union of the qids found in qrels and topics.             |
 +----------------------+----------------------+------------------+--------------------------------------------------------------------+
-| ``False``            | ``True``             | ``A,B``          | ``C`` is removed because it does not appear in the qrels.          |
+| ``False``            | ``True``             | ``B,C``          | ``A`` is removed because it does not appear in the qrels.          |
 +----------------------+----------------------+------------------+--------------------------------------------------------------------+
 
-Note that, following IR evaluation conventions, topics that have no relevance judgments (``C`` in the above example)
+Note that, following IR evaluation conventions, topics that have no relevance judgments (``A`` in the above example)
 do not contribute to relevance-based measures (e.g., ``map``), but still contribute to efficiency measures (e.g., ``mrt``).
 As such, aggregate relevance-based measures will not change based on the value of ``filter_by_qrels``. When ``perquery=True``,
-topics that have no relevance judgments (``C``) will give a value of ``NaN``, indicating that they are not defined
+topics that have no relevance judgments (``A``) will give a value of ``NaN``, indicating that they are not defined
 and should not contribute to the average.
 
 The defaults (``filter_by_topics=True`` and ``filter_by_qrels=False``) were chosen because they likely reflect the intent
--- a/docs/experiments.rst
+++ b/docs/experiments.rst
@@ -147,6 +147,53 @@ This provides a dataframe where each row is the performance of a given system fo
 
 NB: For brevity, we only show the top 5 rows of the returned table.
 
+Missing Topics and/or Qrels
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


pyterrier/pipelines.py
```diff
@@ -147,33 +161,27 @@ def _run_and_evaluate(
...
+            results.append(res)
             starttime = timer()
-        if not perquery:
-            aggregators = {rev_mapping.get(m, str(m)): m.aggregator() for m in metrics}
-            for q in evalMeasuresDict:
-                for metric in metrics:
-                    s_metric = rev_mapping.get(metric, str(metric))
-                    aggregators[s_metric].add(evalMeasuresDict[q][s_metric])
-            evalMeasuresDict = {m: agg.result() for m, agg in aggregators.items()}
+        evalMeasuresDict = _ir_measures_to_dict(
```
seanmacavaney, line 178: This change may be controversial. When `batch_size` is set, it no longer evaluates after each batch, but rather keeps all the results and evaluates everything at the end.

Alternatively, I could update so it filters the qrels each time.

pyterrier/pipelines.py
```diff
@@ -95,7 +97,17 @@ def _ir_measures_to_dict(
             metric = m.measure
             metric = rev_mapping.get(metric, str(metric))
             rtr[m.query_id][metric] = m.value
+        if backfill_qids is not None:
```
seanmacavaney, line 100: It was necessary to move the backfilling stuff down to this level where rev_mapping is available. It was actually buggy before: If a family name was given and perquery was set, the family name would appear with a 0 in the summary, rather than the member names. Tests catch this now.

pyterrier/pipelines.py
```diff
@@ -56,15 +57,15 @@ def _color_cols(data, col_type,
 }
 
 def _convert_measures(metrics : MEASURES_TYPE) -> Tuple[Sequence[BaseMeasure], Dict[BaseMeasure,str]]:
-    from ir_measures import convert_trec_name
+    from ir_measures import parse_trec_measure
```
seanmacavaney, line 60: same function, old name was deprecated

pyterrier/pipelines.py
```diff
@@ -346,6 +354,8 @@ def _apply_round(measure, value):
             evalDict[name] = evalMeasuresDict
         else:
             import builtins
+            if mrt_needed:
```
seanmacavaney, line 357: mrt only applis when perquery is false

cmacdonald:
sorry, evaluation after each `batch_size` is definitely preferred - if you have lots of queries, this avoids keeping ALL of them in memory. Esp. a big challenge for MSMARCO dev/eval sets.

seanmacavaney:
yep, makes sense, I'll update.

cmacdonald (changes requested):

pyterrier/pipelines.py
```diff
@@ -214,6 +223,7 @@ def Experiment(
             Applying a batch_size is useful if you have large numbers of topics, and/or if your pipeline requires large amounts of temporary memory
             during a run.
         drop_unused(bool): If True, will drop topics from the topics dataframe that have qids not appearing in the qrels dataframe. 
+        filter_qrels(bool): If True, will drop topics from the qrels dataframe that have qids not appearing in the topics dataframe.
```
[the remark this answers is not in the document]
seanmacavaney: Yeah, you can have both true. It's essentially left outer join, right outer join, and inner join.

If qrels include A, B and topics include B, C:
 - `drop_unused=True filter_qrels=True` gives B
 - `drop_unused=True filter_qrels=False` gives A, B
 - `drop_unused=False filter_qrels=True` gives B, C
 - `drop_unused=False filter_qrels=False` gives A, B, C

I'm in favour of giving them more similar names. `filter_by_topics` (formerly `filter_qrels`) and `filter_by_qrels` (formerly `drop_unused`) sound reasonable to me. Should probably pull from kwargs and give a warning if `drop_unused` is used for backward compatibility.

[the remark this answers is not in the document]
cmacdonald: Agreed. We may need something in the documentation about this filtering.

Are we sure 
 * `filter_by='topics'`
 * `filter_by='qrels'`
 * `filter_by='both'`
 * `filter_by=None`

shouldnt be the correct nomenclature?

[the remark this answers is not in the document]
seanmacavaney: I've been thinking about it in terms of separate decisions the user could make, and as a result, a single option that does both could be confusing.

1. Do you want skip topics that do not appear in the qrels? If so, set `filter_by_qrels=True`.
2. Do you want to evaluate across all topics from the qrels, even if they do not appear in the requested topics? If so, set `filter_by_topics=False`. I'd expect this to be particularly uncommon and be limited to situations like the one brought up in #226.

I could probably be convinced otherwise, though.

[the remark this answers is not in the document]
cmacdonald: Ok, you convinced me. Carry on.

docs/experiments.rst
```diff
@@ -147,6 +147,53 @@ This provides a dataframe where each row is the performance of a given system fo
...
+| ``True`` (default)   | ``False`` (default)  | ``B,C``          | ``A`` is removed because it does not appear in the topics.         |
++----------------------+----------------------+------------------+--------------------------------------------------------------------+
+| ``True`` (default)   | ``True``             | ``B``            | Acts as an intersection of the qids found in the qrels and topics. |
++----------------------+----------------------+------------------+--------------------------------------------------------------------+
+| ``False``            | ``False`` (default)  | ``A,B,C``        | Acts as a union of the qids found in qrels and topics.             |
++----------------------+----------------------+------------------+--------------------------------------------------------------------+
+| ``False``            | ``True``             | ``A,B``          | ``C`` is removed because it does not appear in the qrels.          |
++----------------------+----------------------+------------------+--------------------------------------------------------------------+
+
```
cmacdonald, line 177: Would it be more intuitive for topics to have A,B and qrels to have B,C, as thats the order in the table.

PR #247, opened 2021-12-08
save_dir for pt.Experiment #163

cmacdonald:


cmacdonald:
Test cases will be added later

pyterrier/io.py
```diff
@@ -212,22 +213,23 @@ def write_results(res, filename, format="trec", **kwargs):
...
-    return SUPPORTED_RESULTS_FORMATS[format][1](res, filename, **kwargs)
+    return SUPPORTED_RESULTS_FORMATS[format][1](res, filename, append=append, **kwargs)
 
-def _write_results_trec(res, filename, run_name="pyterrier"):
+def _write_results_trec(res, filename, run_name="pyterrier", append=False):
         res_copy = res.copy()[["qid", "docno", "rank", "score"]]
         res_copy.insert(1, "Q0", "Q0")
         res_copy.insert(5, "run_name", run_name)
-        res_copy.to_csv(filename, sep=" ", header=False, index=False)
+        res_copy.to_csv(filename, sep=" ", mode='wa' if append else 'w', header=False, index=False)
```
seanmacavaney, line 222: I think these need to just be `'a'` (instead of `'wa'`)

pyterrier/io.py
```diff
@@ -212,22 +213,23 @@ def write_results(res, filename, format="trec", **kwargs):
...
         res_copy = res.copy()[["qid", "docno", "rank"]]
-        res_copy.to_csv(filename, sep="\t", header=False, index=False)
+        res_copy.to_csv(filename, sep="\t", mode='wa' if append else 'w', header=False, index=False)
 
-def _write_results_letor(res, filename, qrels=None, default_label=0):
+def _write_results_letor(res, filename, qrels=None, default_label=0, append=False):
     if qrels is not None:
         res = res.merge(qrels, on=['qid', 'docno'], how='left').fillna(default_label)
-    with autoopen(filename, "wt") as f:
+    mode='wta' if append else 'wt'
```
seanmacavaney, line 231: similar `'at'` not `'wta'`

pyterrier/pipelines.py
```diff
@@ -186,6 +202,11 @@ def _run_and_evaluate(
                 raise ValueError("batch of %d topics, but no results received in batch %d from %s" % (len(batch_topics), i, str(system) ) )
             endtime = timer()
             runtime += (endtime - starttime) * 1000.
+
+            # write results to save_file; we will append for subsequent batches
+            if save_file is not None:
```
seanmacavaney, line 207: should this file be cleaned up if an exception is thrown? Otherwise, the partial file will still be around and picked up on a subsequent execution (mistakenly thinking it's a complete file).

pyterrier/pipelines.py
```diff
@@ -365,6 +388,14 @@ def _apply_round(measure, value):
     elif len(names) != len(retr_systems):
         raise ValueError("names should be the same length as retr_systems")
 
+    # validate save_dir and resulting filenames
+    if save_dir is not None:
+        if not os.path.exists(save_dir):
+            raise ValueError("save_dir %s does not exist" % save_dir)
+        for n in names:
```
seanmacavaney, line 395: should we need to check for duplicate names?

pyterrier/pipelines.py
```diff
@@ -365,6 +388,14 @@ def _apply_round(measure, value):
...
+            if '/' in n:
```
seanmacavaney, line 396: This turns out to be a little more complicated for Windows, since there are more reserved path characters: https://stackoverflow.com/questions/1033424/how-to-remove-bad-path-characters-in-python

pyterrier/pipelines.py
```diff
@@ -365,6 +388,14 @@ def _apply_round(measure, value):
     elif len(names) != len(retr_systems):
         raise ValueError("names should be the same length as retr_systems")
 
+    # validate save_dir and resulting filenames
+    if save_dir is not None:
+        if not os.path.exists(save_dir):
```
seanmacavaney, line 393: Any reason we can't make the path for them if it doesn't exist? I guess there's a risk it could create tons of subdirectories.

Should also check `isdir` instead, otherwise they could get an error later in the off-chance it's a file or something else.

pyterrier/pipelines.py
seanmacavaney: Also, won't appending cause problems for gzip files?

pyterrier/io.py
cmacdonald: Default mode for pandas is 'w', so the documentation said

pyterrier/pipelines.py
cmacdonald: Yes, a catch clause should be added to remove the file.

I can confirm that GZip files can be appended with gzip files.

pyterrier/pipelines.py
cmacdonald: Perhaps I extract a pt.io.ok_filename() method?

pyterrier/io.py
seanmacavaney: My point is that `'wa'` isn't a valid file mode. E.g.,

```python
pd.DataFrame().to_csv('test.csv', mode='wa')
# ValueError: must have exactly one of create/read/write/append mode
```

pyterrier/pipelines.py
cmacdonald: > Any reason we can't make the path for them if it doesn't exist? I guess there's a risk it could create tons of subdirectories.

Hmmm. Not sure. Lets leave as is for now?

> Should also check `isdir` instead, otherwise they could get an error later in the off-chance it's a file or something else.
I added this.

pyterrier/pipelines.py
```diff
@@ -181,25 +197,36 @@ def _run_and_evaluate(
...
+                evalMeasuresDict.update(_ir_measures_to_dict(
+                    ir_measures.iter_calc(metrics, batch_qrels, res.rename(columns=_irmeasures_columns)),
+                    metrics,
+                    rev_mapping,
+                    num_q,
+                    perquery=True,
+                    backfill_qids=batch_backfill))
+                pbar.update()
+                starttime = timer()
+        except Exception as inst:
```
seanmacavaney, line 225: I think this can be simplified to:

```python
except:
    # [remove stuff]
    raise
```

cmacdonald:
ok, what about the names `save_dir`, `save` ="overwrite" or save="reuse" ? I kind of think the latter should be save_mode, but it seems unnecessary

seanmacavaney:
> ok, what about the names `save_dir`, `save` ="overwrite" or save="reuse" ? I kind of think the latter should be save_mode, but it seems unnecessary

"save" could be confused with a boolean argument that just indicates whether it should save or not. Probably safer to use "save_mode".

cmacdonald:
added documentation, can you review that too, then this is done.

cmacdonald:
(failures are unrelated)

PR #249, opened 2021-12-09
use jnius 1.4

cmacdonald:


cmacdonald:
This fixes PyTerrier support for Jnius 1.4. Jnius 1.4 appeared to change some GC behaviour, which caused an error that occasionally appeared on Windows to always occur. I think this fixes the issue.

The benefit of this /should/ be support for Python 3.10.

Questions for consideration during review:
 - should we test python 3.10 in this PR too
 - how should we peg the jnius version: 1.3 -- 1.4? or just >=1.3.0?

cmacdonald:
> should we test python 3.10 in this PR too
separate thing.

> how should we peg the jnius version: 1.3 -- 1.4? or just >=1.3.0?

my preference is now >=1.4.2. We dont test jnius version in github actions, so lets force to the latest. we may restrict to 1.4.x later.

PR #254, opened 2021-12-20
Set dtype for qrels columns at read time in io method

jjdelvalle:
The current approach allows pandas to infer the type of each column at read time and then the types for the `qid` and `docno` columns are forced to be `str`. This could result in unexpected behavior if one of those fields is number-like with leading zeros.

This would make it so a `docno` or `qid` such as `0001` is read as `1` by pandas, and then casted to `"1"`.

This change makes it so, pandas gets the type explicitly and the field can be read directly as `"0001"`, thus conserving the leading zeros. 

(I also removed some trailing space in the source code)


cmacdonald:
Good spot, and thanks for this PR. I updated a corresponding test case.

jjdelvalle:
Good thinking about the test case and thanks for accepting!

PR #256, opened 2021-12-20
faster testing in Github Actions: focus on requested jnius, rather than changing jnius version 3 times

cmacdonald:


cmacdonald:
Shows to be faster.

PR #258, opened 2021-12-20
No coerce, pt.Transformer

cmacdonald:


pyterrier/transformer.py
```diff
@@ -244,52 +262,55 @@ def set_parameter(self, name : str, value):
...
 
-    def __invert__(self):
+    def __invert__(self : 'Transformer') -> 'Transformer':
         from .cache import ChestCacheTransformer
         return ChestCacheTransformer(self)
 
     def __hash__(self):
         return hash(repr(self))
 
+class TransformerBase(Transformer):
```
seanmacavaney, line 311: Could this be aliased as `TransformerBase = Transformer` instead? That way, things that inherit from it will be _exactly_ the same as if they inherit from `Transformer`.

(A downside is that it wouldn't allow us to add deprecation warnings to it.)

Maybe a quick comment here too to say that it's to support an old convention.

seanmacavaney (approved):
love it, just one minor comment

pyterrier/transformer.py
cmacdonald: Thanks. I think I'll go for the comment, and then later the deprecation warning in v0.9.

PR #260, opened 2021-12-22
pyterrier loaded message to stderr

seanmacavaney:
When using pyterrier in a script and redirecting stdout (e.g., to a file), it's inconvenient to have the "PyTerrier has loaded Terrier" message appear at the top.


cmacdonald (approved):

cmacdonald:
Fair point.

PR #261, opened 2022-01-06
Fix code block in `ltr.rst` in section `Working with Features`

bart-kosmala:
Small typos, but they make it confusing while reading the chapter.


cmacdonald (approved):

cmacdonald:
Good spot @bart-kosmala. Thanks for the PR!

PR #262, opened 2022-01-11
provide re-ranking runs from datasets

seanmacavaney:
It's becoming increasingly common for datasets to include a standard initial result set for re-ranking experiments. ir_datasets provides these, but they are a bit inconvenient to use in pyterrier right now because it involves using `irds_ref`, building an appropriate dataframe, and merging in the query/document text.

This PR exposes the functionality more conveniently through `dataset.get_reranking_run()`. (I'm open to alternative names here.)

I also considered including the document text, but it's not always going to be needed and it seems more suitable to use the existing `pt.text.get_text()` pipeline for this.

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -107,6 +107,12 @@ class Dataset():
         """
         return None
 
+    def get_reranking_run(self, variant=None) -> pd.DataFrame:
+        """ 
+            Returns a standard run provided by the dataset used for re-ranking experiments.
+        """
+        pass
+
 class RemoteDataset(Dataset):
 
     def __init__(self, name, locations):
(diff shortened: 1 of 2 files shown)


seanmacavaney:
If this looks good, I'll add documentation for it too.

cmacdonald:
Thanks. A smaller name would be appreciated. Why not just `get_run()`?

We need to discuss timing. I wasnt planning to release 0.8 until March, i.e. after IR(H)

Can we fix #212 in 0.8 as well.

cmacdonald:
or `get_results()`? As I dont think we use the word run in PyTerrier.

seanmacavaney:
`get_results` seems a bit generic, but it'll probably work.

I don't think this urgently needs to hit pypi. Sorry I lost track of #212-- I'll fix it as well.

cmacdonald (approved):

PR #265, opened 2022-01-11
read_results merge topics

seanmacavaney:
When reading results files for further processing (e.g., using a saved result file for further re-ranking), you usually will need an additional column that's not saved in the file: `query`. This can be accomplished by the user by loading the dataset's topics and performing a `pd.merge()` on the results and topics dataframes, but this is inconvinenet.

This PR provides several options to help the user load a results dataframe that includes topics with more ease. They can either:
 - Pass a `dataset` into `read_results`, either as a string identifier or as a Dataset object. The topics for this dataset will be loaded and merged with the results.
 - Pass `topics` into `read_results`. These topics will be merged with the results. This is handy if the user already has the topics handy, or if they do not come directly from a dataset.

Functionality would be ambiguous if both options were given, so these two options are exclusive. An assertion error will be thrown if both are provided by the user.

### DIFF
--- a/pyterrier/io.py
+++ b/pyterrier/io.py
@@ -148,6 +148,27 @@ def read_results(filename, format="trec", topics=None, dataset=None, **kwargs):
 
     Returns:
         dataframe with usual qid, docno, score columns etc
+
+    Examples::
+
+        # a dataframe of results can be used directly in a pt.Experiment
+        pt.Experiment(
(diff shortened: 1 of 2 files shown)


cmacdonald:
I think we need to add some examples of the various uses of `read_results()`, at least one using dataset and topics options? 
We might want to add another example using `pt.Experiment()`?

PR #272, opened 2022-01-20
Fix PyJnius imports

heinrichreimer:
In the datasets module `pt.autoclass` is imported which doesn't exist.
Replaced it with the direct import from PyJnius.

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -366,12 +366,12 @@ class RemoteDataset(Dataset):
         return None
 
     def get_index(self, variant=None, **kwargs):
-        import pyterrier as pt
+        from jnius import autoclass
         if self.name == "50pct" and variant is None:
             variant="ex1"
         thedir = self._get_all_files("index", variant=variant, **kwargs)
         return thedir
-        #return pt.autoclass("org.terrier.querying.IndexRef").of(os.path.join(thedir, "data.properties"))
+        #return autoclass("org.terrier.querying.IndexRef").of(os.path.join(thedir, "data.properties"))
 
     def __repr__(self):
         return "RemoteDataset for %s, with %s" % (self.name, str(list(self.locations.keys())))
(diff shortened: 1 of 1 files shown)


cmacdonald:
Thanks for the PR @heinrichreimer 

Do you have a test case for this? pt.autoclass should exist as long as pt.init() has been called - see https://github.com/terrier-org/pyterrier/blob/master/pyterrier/__init__.py#L148.

This code has a lot of testing in the unit tests, so I'm comfortable it works.

heinrichreimer:
Ah I see, I havent called `init()`, my bad

cmacdonald:
> Ah I see, I havent called `init()`, my bad

Correct. `init()` is necessary to ensure appropriate JVM configuration. All we can do is put more checks that things aren't called before init().

heinrichreimer:
Or can we just integrate
```python
if not started():
    init()
```
into the methods that need the JVM to be initialized?

heinrichreimer:
Still I'm a little bit confused as `autoclass` seems to be imported directly from `jnius` [in some places](https://github.com/terrier-org/pyterrier/search?q=jnius+autoclass) and from `pyterrier` [in some other places](https://github.com/terrier-org/pyterrier/search?q=pt+autoclass).

cmacdonald:
> Or can we just integrate
> ```python
> if not started():
>     init()
> ```
> into the methods that need the JVM to be initialized?

We've discussed automatic init internally at length. Auto initiation has some dangerous unintended side-effects - for instance, if multiple libraries use Jnius.

I'm not saying this is a hill I'm willing to die on. But its fairly easy to get used to start with pt.init().

> Still I'm a little bit confused as `autoclass` seems to be imported directly from `jnius` [in some places](https://github.com/terrier-org/pyterrier/search?q=jnius+autoclass) and from `pyterrier` [in some other places](https://github.com/terrier-org/pyterrier/search?q=pt+autoclass).

There is/was some intention to replace jnius with jpype at some point, hence the aliasing. However, there is no marked efficiency benefit, so we backburnered that effort.

cmacdonald:
In your code, you could `assert pt.started()` at the top of the relevant file, e.g. https://github.com/heinrichreimer/chatnoir-pyterrier/blob/main/chatnoir_pyterrier/retrieve.py

heinrichreimer:
> In your code, you could `assert pt.started()` at the top of the relevant file, e.g. https://github.com/heinrichreimer/chatnoir-pyterrier/blob/main/chatnoir_pyterrier/retrieve.py

Very good idea! I'm closing this PR as it's not really needed aymore.

PR #274, opened 2022-01-23
fixed bug with is_transformer

seanmacavaney:
Related to #258 -- subclasses of `pt.Transformer` doesn't pass `is_transformer` check because it checks for `TransformerBase` (a subclass of `Transformer`).

This means that you wouldn't be able to use a subclass of `pt.Transformer` in a pipeline. E.g.,

```python
class MyModel(pt.Transformer):
    pass
bm25 >> MyModel() # would fail with message saying MyModel cannot be coerced into a transformer
```

Since most code still uses `TransformerBase`, it doesn't break existing code -- just new stuff written using the new name.

Added tests that failed with previous impl and added fix.

### DIFF
--- a/pyterrier/transformer.py
+++ b/pyterrier/transformer.py
@@ -16,7 +16,7 @@ def is_function(v):
     return isinstance(v, types.FunctionType)
 
 def is_transformer(v):
-    if isinstance(v, TransformerBase):
+    if isinstance(v, Transformer):
         return True
     return False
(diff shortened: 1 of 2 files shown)


seanmacavaney:
One of the checks failed, but it just seems to be a network error.

cmacdonald:
yes. enough passed that I was comfortable

PR #280, opened 2022-02-28
defer loading ir_datasets

seanmacavaney:
Context: calling `ir_datasets.load` can give a warning message that indicates if a dataset is deprecated. Without this change, the warning would come up every time, regardless of whether the dataset was actually used. With the change, the warning will only appear if the user specifically requests the dataset and does something with it (only the first time).

This change does not degrade the behaviour of on-demand generated datasets (such as those from CLIRMatrix), as the load is not deferred in that cause.

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -381,11 +381,13 @@ class RemoteDataset(Dataset):
 
 
 class IRDSDataset(Dataset):
-    def __init__(self, irds_id):
+    def __init__(self, irds_id, defer_load=False):
         self._irds_id = irds_id
-        self._irds_ref = ir_datasets.load(self._irds_id)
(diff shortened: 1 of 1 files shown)


cmacdonald:
Doesnt this undo https://github.com/terrier-org/pyterrier/pull/263?

seanmacavaney:
Nope -- deferred loading only happens for datasets explicitly in the registry -- which are by definition existent. On-demand generated datasets (such as those from CLIRMatrix) are loaded as they are requested, so it'll throw an error if it doesn't exist.

It looks like the failed tests are due to the current lapse of availability of the ANTIQUE collection -- see here: https://github.com/seanmacavaney/ir-datasets.com/runs/5363947696

cmacdonald:
OK. 

I think CIIR.umass.edu is just slow, but its currently up.

PR #284, opened 2022-03-29
fix LTR groupby for xgboost & lightgbm

cmacdonald:


cmacdonald:
test failure now addressed

pyterrier/ltr.py
```diff
@@ -128,9 +128,9 @@ def fit(self, topics_and_results_Train, qrelsTrain, topics_and_results_Valid, qr
         kwargs = self.fit_kwargs
         self.learner.fit(
             np.stack(tr_res["features"].values), tr_res["label"].values, 
-            group=tr_res.groupby(["qid"]).count()["docno"].values, # we name group here for lightgbm compat. 
+            group=tr_res.groupby(["qid"])["docno"].count().to_numpy(), # we name group here for lightgbm compat.
```
seanmacavaney, line 131: If I properly understand how `.fit()` works, `tr_res` and `va_res` will need to be sorted by `qid` for this to work as expected, since `.groupby()` sorts by the key.

This isn't a degradation from the current implementation though.

seanmacavaney (approved):

pyterrier/ltr.py
cmacdonald: Hmm interesting point. `tr_res` and `va_res` are `pd.merge`d with the qrels which I think will force the ordering.

cmacdonald:
What Sean had identified was indeed a bug affecting xgboost or lightgbm integration. In training or validation sets with differing number of results per query, hence documents could be associated to the wrong query by the learner. This fixes the issue through sorting of the resultset, and introduces a new unit test to check the issue.

In most LTR scenarios, there will be 1000 docs per query, but I'll examine differences in a realistic LTR setting.

pyterrier/ltr.py
```diff
@@ -112,6 +112,13 @@ def fit(self, topics_and_results_Train, qrelsTrain, topics_and_results_Valid, qr
             qrelsValid(DataFrame): A dataframe containing the qrels for the validation topics
             
         """
+
+        def _count_by_topic(res : pd.DataFrame) -> Sequence[int]:
```
seanmacavaney, line 116: type annotation should be Tuple[Sequence[int], pd.DataFrame]?

seanmacavaney (approved):

PR #303, opened 2022-05-05
fix error when a query field is named "query"

### LINKED ISSUE #302: irds topics variants throughs error
**Describe the bug**
```
pt.get_dataset('irds:beir/trec-covid').get_topics('text') 
```

seanmacavaney:
Some datasets have a query field called "query". We need to remove it or we'll end up with multiple "query" columns, which will cause problems because many components are written assuming no columns have the same name.

This only affects a few datasets. I went with `irds:clueweb09/catb/trec-web-2009` in the test because it only needs to download the topics file (compared to BEIR, which would need to download an archive that contains the corpus, etc.) Plus, the topic file is mirrored, so we'll probably not encounter issues arising from intermittent network failures.

fixes #302

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -431,6 +431,11 @@ class IRDSDataset(Dataset):
         df.rename(columns={"query_id": "qid"}, inplace=True) # pyterrier uses "qid"
 
         if variant is not None:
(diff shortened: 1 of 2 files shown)


cmacdonald (approved):

cmacdonald:
Nice fast job Sean

PR #305, opened 2022-05-15
faster version of ConcatenateTransformer

seanmacavaney:
`ConcatenateTransformer` (i.e., the `^` operator) is _really_ slow for large result sets.

This patch should be functionally identical to the previous implementation but is substantially faster. It eliminates all the messing around with pandas `merge`s in favor of working over each query one-at-a-time. It also avoids building temporary columns since `first_score` and `last_score` are the same for the entire query.

### DIFF
--- a/pyterrier/transformer.py
+++ b/pyterrier/transformer.py
@@ -514,30 +514,43 @@ class ConcatenateTransformer(BinaryTransformerBase):
         import pandas as pd
         # take the first set as the top of the ranking
         res1 = self.left.transform(topics_and_res)
-        # identify the lowest score for each query
-        last_scores = res1[['qid', 'score']].groupby('qid').min().rename(columns={"score" : "_lastscore"})
 
         # the right hand side will provide the rest of the ranking        
         res2 = self.right.transform(topics_and_res)
 
-        
-        intersection = pd.merge(res1[["qid", "docno"]], res2[["qid", "docno"]].reset_index())
-        remainder = res2.drop(intersection["index"])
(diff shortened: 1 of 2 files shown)


seanmacavaney:
Just noticed that there isn't a test for `ConcatenateTransformer`. I can look into adding one when I have more time.

cmacdonald:
> Just noticed that there isn't a test for `ConcatenateTransformer`. I can look into adding one when I have more time.

See https://github.com/terrier-org/pyterrier/blob/master/tests/test_operators.py#L184-L201 ?

cmacdonald:
ping on the first comment - would like to merge soon

seanmacavaney:
Hmm, I'm no longer seeing the performance gains by using this implementation, and the current implementation doesn't take probability long anymore. Possibilities: different Pandas version than before, something about my benchmark is different, something else (?).

cmacdonald:
you would expect the Pandas version to be parallelise or better in native...

seanmacavaney:
Often yes, but that's not what I saw in this case. I'll have to go back and try to figure out the precise pipelines I was running. But I'm certain that one with pandas merge was running too slow to use in my setting -- which I why I spent the time implementing a faster version.

cmacdonald:
One query vs many queries might make a difference

seanmacavaney:
Yes, and the benchmark I wrote yesterday tested with many queries.

seanmacavaney:
closing because I was unable to reproduce the original issue this PR was meant to solve

PR #313, opened 2022-07-25
Fix type annotation

### LINKED ISSUE #312: Incorrect type annotation for "cutoff" operator
Should be `int` instead of `Transformer`. 
https://github.com/terrier-org/pyterrier/blob/943764f32fc04ae0f164f3ec1a399cd28ea8e94b/pyterrier/transformer.py#L297-L299

Expressions such as the following would otherwise trigger a type check warning:
```python
duo_pipeline = mono_pipeline % 50 >> duoT5
```

heinrichreimer:
fixes #312


seanmacavaney:
Thanks -- good spot and lgtm.

PR #286, opened 2022-04-01
Api fixups

cmacdonald:

PR #314, opened 2022-07-27
0.9refactor

cmacdonald:
this looks similar to #286. TODO see whats here, and whats not.


seanmacavaney:
Considering other plugin libraries that currently reference `TransformerBase` instead of `Transformer`, I'm not sure we can drop it quite yet.

cmacdonald:
its still there...!

seanmacavaney:
Ah, okay, sorry. The diff makes it look like it was removed.

cmacdonald:
Just need to sort out indexing API, like in #286

cmacdonald:
The advantages of this PR:

We now have stable public APIs for the main transformer classes:
 pt.Transformer, pt.Estimator, pt.Indexer

Older classes have been deprecated (pt.TransformerBase, pt.EstimatorBase, pt.IterDictIndexerBase)

Refactoring as follows:
 - pyterrier/transformer.py is now smaller and focused on the public API imple 
 - pyterrier/ops.py contains the operator transformer classes only (compose, feature union etc,)
 - pyterrier/apply.py - apply classes have been moved from transformer.py into pyterrier/apply_base.py

seanmacavaney (approved):

seanmacavaney:
The code changes look good to me. A few suggestions:
 - Let's run tests on some of the plugin libraries, just to be sure we're not missing something major in the interface.
 - Can we write a list of breaking changes? E.g., stuff that moved from one namespace to another? They don't seem like things folks are likely to be referenced directly, but just in case. We can include it in the release notes.

cmacdonald:
Agreed to all comments.

PR #320, opened 2022-08-05
Argument meta should be supplied as dictionary

JohnGiorgi:
[The documentation](https://pyterrier.readthedocs.io/en/latest/terrier-indexing.html#pyterrier.IterDictIndexer) specifies the `meta` argument to `pt.IterDictIndexer` as a dictionary mapping meta field names to their lengths, e.g. 

```python
iter_indexer = pt.IterDictIndexer("./index", meta={'docno': 20, 'text': 4096})
```

but the documentation _examples_ break this into two list args, `meta` and `meta_lengths`

```
iter_indexer = pt.IterDictIndexer("./index", meta=['docno', 'text'], meta_lengths=[20, 4096])
```

I think this is a mistake? I am finding that my length arguments are not respected when passed similar to the documentation examples.

### DIFF
--- a/docs/terrier-indexing.rst
+++ b/docs/terrier-indexing.rst
@@ -89,7 +89,7 @@ IterDictIndexer
 An iterdict can just be a list of dictionaries::
 
     docs = [ { 'docno' : 'doc1', 'text' : 'a b c' }  ]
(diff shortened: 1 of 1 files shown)


cmacdonald:
Good spot, thanks @JohnGiorgi

PR #322, opened 2022-08-19
Python 3.10 support

cmacdonald:


cmacdonald:
@seanmacavaney 

I have released pytrec_eval_terrier 0.5.3. but ir_measures is pegged to a particular version 

```
The conflict is caused by:
    The user requested pytrec_eval_terrier>=0.5.3
    ir-measures 0.3.0 depends on pytrec-eval-terrier==0.5.2
```

Could ir-measures be a bit less fussy?

cmacdonald:
documentation of changes:
 - Chest has had no release compatible with Python 3.9. I think we can remove Chest in the future, but for now, lets install from Github
 - Some Python Windows images dont have Torch installed - lets separate and skip the test without Torch installed
 - Use modern version of Pytrec_eval and ir-measures.

NB: No Jnuis binaries for Python 3.11 yet, so newest support is 3.10.

.github/workflows/push.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     
     strategy:
       matrix:
-        python-version: ['3.7', '3.9']
+        python-version: ['3.7', '3.10']
```
seanmacavaney, line 17: Any reason not to list all of 3.7, 3.8, 3.9, 3.10?

seanmacavaney: Ah, perhaps because each one adds 6 more tests for java/os combinations?

requirements.txt
```diff
@@ -6,7 +6,7 @@ pyjnius>=1.4.2
 matchpy
 sklearn
 deprecation
-chest
+git+https://github.com/s-t-e-v-e-n-k/chest.git@python-310-support#egg=chest
```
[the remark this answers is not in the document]
cmacdonald: One monkeypatching alternative could be:
```python
import collections
import collections.abc
collections.MutableMapping = collections.abc.MutableMapping
```

.github/workflows/push.yml
cmacdonald: Yes. I think we only need to test the extremes.

seanmacavaney (approved):

Issue #321, opened 2022-08-07
Apply doc_score to batches of documents


JohnGiorgi:
**Is your feature request related to a problem? Please describe.**

I am wondering if it is possible to apply the `doc_score` transformer in _batches_. The motivation is that this appears to be the primary interface for adding custom re-rankers, and many of these re-rankers will be neural language models whose inference speed can be dramatically increased by batching. Currently, the `fn` provided to `doc_score` is applied to each row of the dataframe independently.

**Describe the solution you'd like**

An argument, e.g. `batch_size` to the `doc_score` transformer that controls how many rows in the `pd.Series` object the provided `fn` receives.

**Describe alternatives you've considered**

- Approach 1: Add an argument `batch_size` to `doc_score`. This could be passed to `ApplyDocumentScoringTransformer`, which would be used in `transformer` to chunk the `inputRes` dataframe into `batch_size` chunks, provding each to `fn` in a loop and collecting the `scores`.
- Approach 2: Add an argument  to `doc_score`, say `all_rows`, which if positive passes the entire dataframe to `fn`. Batching would then be up to the `fn` implementation.

**Additional context**

I would be happy to open a PR if either approach seems promising, and this is a feature that would be welcome in the library. Approach 2 seems simpler and would allow the caller to implement batching however they want. Would look something like:

```python
import warnings
from pyterrier.apply import ApplyTransformerBase
from pyterrier.model import add_ranks

class ApplyDocumentScoringTransformer(ApplyTransformerBase):
    """
        Implements a transformer that can apply a function to perform document scoring. The supplied function 
        should take as input one row, and return a float for the score of the document.
        
        Usually accessed using pt.apply.doc_score()::
            def _score_fn(row):
                return float(row["url".count("/")])
            
            pipe = pt.BatchRetrieve(index) >> pt.apply.doc_score(_score_fn)
    """
    def __init__(self, fn,  *args, **kwargs):
        """
            Arguments:
             - fn (Callable): Takes as input a panda Series for a row representing that document, and returns the new float doument score 
        """
        super().__init__(fn, *args, **kwargs)
    
    def transform(self, inputRes, all_rows = False):
        fn = self.fn
        outputRes = inputRes.copy()
        if all_rows:
            if self.verbose:
                warnings.warn("Cannot provide progess when all_rows is True")
            outputRes["score"] = fn(outputRes)
        else:
            if self.verbose:
                tqdm.pandas(desc="pt.apply.doc_score", unit="d")
[...]

PR #325, opened 2022-08-26
addresses #321 pt.apply.doc_score batching

cmacdonald:


cmacdonald:
Disadvantage with this version: batches can fall across qids. This means that add_ranks() has to be executed on the entire returned dataframe

JohnGiorgi:
I think a slight disadvantage to this is that the caller might have their own batching strategy. 

For example, say you wanted to use `apply_doc_score` with a [Sentence Transformer](https://www.sbert.net/) model either as a retriever or a re-ranker (which I imagine would be a popular use case). Sentence Transformers has a smart batching strategy that batches text of similar length together, so to gain any benefit from that, you would now need to set two batch sizes, `pyterrier_batch_size` >> `sentence_transformers_batch_size`.

One solution would be to add another argument, `batched` (`False` by default), similar to [HF Datasets](https://huggingface.co/docs/datasets/v2.4.0/en/package_reference/main_classes#datasets.Dataset.map.batched), so that:

- If `batched`, then `batch_size` is considered. Otherwise, this is row-wise.
- If `batched` and `batch_size` is `None`, return the whole dataset and let the caller figure out batching.

This is mainly a nitpick though because the same thing could be achieved by setting `batch_size` to the size of the data frame you are applying the transform to.

cmacdonald:
Hi @JohnGiorgi 

We're a bit confused - if the toolkit being called already has batching support, then why need a version of doc_score that performs batching? pt.apply.generic and by_query apply no batching, and could be used when the underlying library does the batching. E.g.

```python

def _scoring_fn(df):
  df["score"] = transformer_library.score_in_own_batches(df["query"], df["text"])

pipe = bm25 >> pt.apply.generic(_scoring_fn)
#pt.apply.by_query could also be used here rather than pt.apply.generic
```

JohnGiorgi:
Okay I think I understand!

My original motivation was to use a Sentence Transformer model as a re-ranker, so I was looking at `doc_score`. But I see your point about using `generic`. The only thing I would need to do (besides implementing the `fn`), is to call `add_ranks`, then>

cmacdonald:
> The only thing I would need to do (besides implementing the fn), is to call add_ranks, then?

yes. Technically add_ranks() is optional, but advised. Applying a rank cutoff with an outdated rank attribute would of course not work.

JohnGiorgi:
Gotcha, okay happy to go this route then. Thanks for the help!

cmacdonald:
Noted. I think the PR still has value, and we can merge it after review.

I think @JohnGiorgi what would be insightful is knowing if the documentation for pt.apply is missing something?

JohnGiorgi:
I think if `pt.apply` is the main way to use custom neural models for retrieval and (re)ranking, it might be helpful to see an example?

cmacdonald:
I added a documentation example for batched `pt.apply.doc_score()`. I think this resolves the request.

PR #343, opened 2022-11-10
Nofifo pretok indexing fixes

cmacdonald:
After Terrier 5.7 release and the helper release to Maven, it became clear that pretokenised wasnt working on Windows. This PR addresses this (its a race-like GC problem actually), and also works around a minor bug in Tr5.7


seanmacavaney (approved):
lgtm

Issue #348, opened 2022-12-06
DFIndexer error messages


maxhenze:
While reproducing the [PyTerrier Indexing Notebook](https://github.com/terrier-org/pyterrier/blob/master/examples/notebooks/indexing.ipynb) a lot of JVM exceptions occur in my notebook. 

Explicitly I'm loading an ir_dataset and try to index: 
```
import ir_datasets
import pandas as pd
import pyterrier as pt

if not pt.started():
  pt.init()

dataset_msmarco_document_trec_dl_2019_judged = ir_datasets.load("msmarco-document/trec-dl-2019/judged")
docstore = dataset_msmarco_document_trec_dl_2019_judged.docs_store()

df_queries = pd.DataFrame(dataset_msmarco_document_trec_dl_2019_judged.queries_iter())
df_qrels = pd.DataFrame(dataset_msmarco_document_trec_dl_2019_judged.qrels_iter())
df_docs = pd.DataFrame(docstore.get_many_iter(df_qrels.doc_id.unique()))

index_path = "./pd_index"
indexer = pt.DFIndexer(index_path, verbose=True)

indexref = indexer.index(df_docs["body"])
```

This results in the following error: 
```

  0%|                                                                               | 0/16043 [00:00<?, ?documents/s]

---------------------------------------------------------------------------
JavaException                             Traceback (most recent call last)
Input In [43], in <cell line: 1>()
----> 1 indexref = indexer.index(df_docs["body"])

File /opt/conda/lib/python3.8/site-packages/pyterrier/index.py:667, in DFIndexer.index(self, text, *args, **kwargs)
    665     javaDocCollection = TQDMSizeCollection(javaDocCollection, len(text)) 
    666 index = self.createIndexer()
--> 667 index.index(autoclass("org.terrier.python.PTUtils").makeCollection(javaDocCollection))
    668 global lastdoc
    669 lastdoc = None

File jnius/jnius_export_class.pxi:1177, in jnius.JavaMultipleMethod.__call__()

File jnius/jnius_export_class.pxi:885, in jnius.JavaMethod.__call__()

File jnius/jnius_export_class.pxi:982, in jnius.JavaMethod.call_method()

File jnius/jnius_utils.pxi:91, in jnius.check_exception()

JavaException: JVM exception occurred: For input string: "" java.lang.NumberFormatException
```

Additionally, if I try to use the indexer as follows: 
```
indexref = indexer.index(df_docs["body"], df_docs["doc_id"])
```
[...]

PR #349, opened 2022-12-06
improve error messages for invalid indexing configurations

cmacdonald:
This addresses the experience reported in #348 by @maxhenze

### DIFF
--- a/pyterrier/index.py
+++ b/pyterrier/index.py
@@ -613,6 +613,9 @@ class DFIndexUtils:
             else:
                 raise ValueError("Keyword kwargs need to be of type pandas.Series, list or tuple")
 
+        if "docno" not in all_metadata:
+            raise ValueError('No docno column specified, while PyTerrier assumes a docno should exist. Found meta columns were %s' % str(list(all_metadata.keys())))
+
         # this method creates the documents as and when needed.
         def convertDoc(text_row, meta_column):
             if text_row is None:
--- a/pyterrier/index.py
+++ b/pyterrier/index.py
@@ -719,7 +719,6 @@ class PythonListIterator(PythonJavaClass):
             lastdoc = [text, meta]
         return lastdoc
 
-
 class FlatJSONDocumentIterator(PythonJavaClass):
     __javainterfaces__ = ['java/util/Iterator']
 
--- a/pyterrier/index.py
+++ b/pyterrier/index.py
@@ -725,7 +725,7 @@ class FlatJSONDocumentIterator(PythonJavaClass):
     def is_dict(obj) -> bool:
         return hasattr(obj, '__getitem__') and hasattr(obj, 'items')
 
-    def __init__(self, it):
+    def __init__(self, it, meta):
         super(FlatJSONDocumentIterator, self).__init__()
         if FlatJSONDocument is None:
             run_autoclass()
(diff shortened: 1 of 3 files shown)


seanmacavaney:
Along this line, a few situations I've seen students encounter recently that could use better indexing error messages:
 - Short circuit error message if a docno is encountered that's longer than the configured maximum. (Currently no error thrown, only know once you get very poor performance when retrieving).
 - Clearer error message when user passes non-dict objects to IterDictIndexer (it's caught in Java and is rather cryptic)

cmacdonald:
> Clearer error message when user passes non-dict objects to IterDictIndexer (it's caught in Java and is rather cryptic)

I have introduced a check for this on the first document. There is a related issue (tracking at https://github.com/terrier-org/terrier-core/issues/215) of not propagating the Java exception back into Python by FLATJsonDocument that we can fix upstream for a future release.

> Short circuit error message if a docno is encountered that's longer than the configured maximum. (Currently no error thrown, only know once you get very poor performance when retrieving).

Actually, this is caused by  `'metaindex.compressed.crop.long' : 'true'` which you have for IterDictIndexer (presumably to make text cropping automatic) - see https://github.com/terrier-org/pyterrier/blob/master/pyterrier/index.py#L803. We could pass the meta lengths to FlatJSONDocumentIterator and check the validity of the first document? This wont handle all possible error modes (e.g. first document is fine, but latter document is not), but would help, right?

seanmacavaney:
Awesome, thanks!

> This wont handle all possible error modes (e.g. first document is fine, but latter document is not), but would help, right?

Yeah, this would probably cover most cases and would be a huge help.

But cases where things are numbered (e.g., `%p1` ... `%p10`) could consistently fail. The error message in the passage case would be that a duplicate docno was found, and the root cause could be tricky to get to the bottom of without this check.

The reasoning for only checking the first one is that checking every docno would get expensive? I imagine that it wouldn't add all that much overhead, especially as compared to the JSON generation, crossing between Python/Java, doing the actual indexing, etc.

cmacdonald:
Ok, I have adjusted FlatJSONDocumentIterator to handle first doc. I note that FlatJSONDocumentIterator is only used by the non-FIFO setting, ie Windows, so these improvements would not work on Linux/macOS. The fifo impl seems to do more things natively in Java, and I cant see easily how to fix this. 

Aside: the fifo impl is more complicated and possibly unnecessary if the default is threads=1.

seanmacavaney:
> and I cant see easily how to fix this

Should be able to just peek at the first doc [here](https://github.com/terrier-org/pyterrier/blob/master/pyterrier/index.py#L950), no? If you like, I could maybe take a look. 

> Aside: the fifo impl is more complicated and possibly unnecessary if the default is threads=1.

I suspect it's far more efficient though. From what I've seen benchmarking jnius, it's costly to move between Python and Java (which the nofifo one needs to do for every document, while the fifo one does not).

cmacdonald:
> If you like, I could maybe take a look.

Sure, go for it.

pyterrier/index.py
```diff
@@ -849,7 +883,7 @@ def index(self, it, fields=('text',), meta=None, meta_lengths=None, threads=None
             #     {'docno' : 'd1', 'toks' : {'a' : 1, 'aa' : 2}}
             # ]
             
-            iter_docs = DocListIterator(it)
+            iter_docs = DocListIterator(self._filter_iterable(it, fields))
```
seanmacavaney, line 886: I'm not sure why `self._filter_iterable` wasn't here before, as it is in the other invocation below.

seanmacavaney:
@cmacdonald -- updated! Mind taking a look when you have a chance.

I also changed the warning to an error in the case of docno. I feel that if your docnos are being truncated, it is always a problem and it's best to force the user to deal with it before indexing will start. Other fields do not matter as much, so they remain just a warning.

cmacdonald:
lgtm, after discussion

PR #358, opened 2023-01-04
stringify properties and controls, addresses #357

### LINKED ISSUE #357: Cannot set term.max.length property
**Describe the bug**
I seem not to be able to set `term.max.length` property. After I run the code below an error is raised saying:

> ERROR org.terrier.structures.seralization.FixedSizeTextFactory - Term 12265Changed###PubDate:01-01-02;Zone:OC;Ed:1;Section:F;Page:F1;PubCha written in 70 bytes. Max expected size was 61. Increase relevant property: max.term.length for Lexicon, or indexer.meta.forward.keylens for metadata

yet I set the property to higher value.

**To Reproduce**
Unfortunately the data I work with are not open sources, so I cannot share them here. But the minimal setup would be something like this:

```python
pt.set_property("max.term.length", 80)
pt.set_property("indexer.meta.forward.keylens", 80)
index = pt.TRECCollectionIndexer(
    "./trec_index",
    threads=1,
    stemmer=pt.index.TerrierStemmer.none,
    stopwords=pt.index.TerrierStopwords.none,
    tokeniser=pt.index.TerrierTokeniser.whitespace,
    overwrite=True,
    properties={"max.term.length": 80, "indexer.meta.forward.keylens": 80},
)


index_ref = index.index(en_doc_paths[:1])
```

I am not sure where to set the above properties as I was unable to find it in the docs. So I just set it wherever I've seen properties set throughout the documentation and Github issues.

**Expected behavior**
An error should not be thrown or the error should have an updated value for `max.term.length` property.

**Documentation and Issues**
 - [x] I have checked the [PyTerrier documentation](https://pyterrier.readthedocs.io/en/latest/) for relevant content
 - [x] I have checked for previous relevant [PyTerrier issues](https://github.com/terrier-org/pyterrier/issues)

**Additional context**
I am new to Terrier and PyTerrier so I may be missing something obvious. However I am stuck on this for the last 4 hours.

cmacdonald:
closes #357

### DIFF
--- a/pyterrier/__init__.py
+++ b/pyterrier/__init__.py
@@ -319,15 +319,15 @@ def set_property(k, v):
         methods or arguments. So this method should be seen as a safety-valve - a way to override the 
         Terrier configuration not explicitly supported by PyTerrier.
     """
-    properties[k] = v
+    properties[str(k)] = str(v)
     ApplicationSetup.bootstrapInitialisation(properties)
 
 def set_properties(kwargs):
     """
         Allows to set many properties in Terrier's global properties configuration
     """
-    for control, value in kwargs.items():
-        properties.put(control, value)
+    for key, value in kwargs.items():
+        properties[str(key)] = str(value)
     ApplicationSetup.bootstrapInitialisation(properties)
 
 def run(cmd, args=[]):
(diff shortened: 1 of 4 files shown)


cmacdonald:
Tests pass. Thanks to @dburian for the report.

PR #373, opened 2023-02-24
add self-hosted action runner for m1 mac

seanmacavaney:


seanmacavaney:
Will take some more fanagling to get the CI script working with hosted runner. Specifically, "Setup Python" doesn't work (and perhaps other actions later in the script).

seanmacavaney:
Seems to work now. The failing test also fails on `master`.

I added m1 wheels to pypi for dependent packages where necessary. Working on `fastrank` (https://github.com/jjfiv/fastrank/issues/52), but for the interim, I've just (globally) installed a suitable version on the machine. It's only required for tests.

cmacdonald:
good job sean. The failing test is on my todo list.

PR #381, opened 2023-03-21
__call__ to direct on type

cmacdonald:
The purpose of this PR is to make transformer(input) work for both iterdicts and dataframes. This involves adjusting the __call__ method in pt.Transformer.

let `t` be a transformer. Then
`t([{'qid' : 'q100'}])`
is equivalent to `t(df)`


seanmacavaney (approved):
lgtm

PR #383, opened 2023-03-21
Update transformer.rst

albertoueda:
I think the dataframe in the last example should be either `res` or `df`.


seanmacavaney (approved):
looks right, good spot!

PR #384, opened 2023-03-23
deprecate pt.Utils

cmacdonald:


pyterrier/utils.py
```diff
@@ -1,11 +1,13 @@
 import pandas as pd
 from collections import defaultdict
+from deprecated import deprecated
 
 
 class Utils:
-
+    # THIS CLASS WILL BE REMOVED the second minor release after 0.9
```
seanmacavaney, line 7: I'm terrified of this change... `pt.Utils` must be used in so many places; not just here, but in extension packages, scripts, etc. Can't we keep it around longer term for backward compatibility?

cmacdonald: happy to go for longer.

PR #387, opened 2023-03-31
added tokenizer as arg for pt.text.sliding

mihirs16:
For the purposes of obtaining passages of a given length, tokenisation is performed by using the `tokenizer` object passed as an argument. The `tokenizer` object must have a `.tokenize(str) -> list[str]` method. By default, tokenisation is performed by splitting on one-or-more spaces, i.e. based on the Python regular expression ``re.compile(r'\s+')``.

Example
```python
from transformers import AutoTokenizer
tok = AutoTokenizer.from_pretrained("bert-base-uncased")
pipe = (pt.BatchRetrieve(index, wmodel="DPH", metadata=["docno", "body"])
      >> pt.text.sliding(length=128, stride=64, prepend_attr=None, tokenizer=tok)
      >> pt.text.scorer(wmodel="DPH")
      >> pt.text.max_passage() )
```

### DIFF
--- a/pyterrier/text.py
+++ b/pyterrier/text.py
@@ -165,10 +165,12 @@ def sliding( text_attr='body', length=150, stride=75, join=' ', prepend_attr='ti
     applying this transformer, docnos are altered by adding '%p' and a passage number. The original scores for each document can be recovered
     by aggregation functions, such as `max_passage()`.
 
-    For the puposes of obtaining passages of a given length, tokenisation is perfomed by using the `tokenizer` object passed as an argument. 
-    The `tokenizer` object must have a `.tokenize(str) -> list[str]` method. By default, tokenisation is perfomed by splitting on one-or-more spaces, 
-    i.e. based on the Python regular expression ``re.compile(r'\s+')``.
-
+    For the puposes of obtaining passages of a given length, the tokenisation can be controlled. By default, tokenisation takes place by splitting
+    on space, i.e. based on the Python regular expression ``re.compile(r'\s+')``. However, more fine-grained tokenisation can applied by passing 
+    an object matching the HuggingFace Transformers `Tokenizer API <https://huggingface.co/docs/transformers/main/en/main_classes/tokenizer#transformers.PreTrainedTokenizer>`_ 
+    as the `tokenizer` kwarg argument. In short, the `tokenizer` object must have a `.tokenize(str) -> list[str]` method and 
+    `.convert_tokens_to_string(list[str]) -> str` for detokenisation.
+    
     Parameters:
         text_attr(str): what is the name of the dataframe attribute containing the main text of the document to be split into passages.
             Default is 'body'.
--- a/pyterrier/text.py
+++ b/pyterrier/text.py
@@ -177,7 +177,7 @@ def sliding( text_attr='body', length=150, stride=75, join=' ', prepend_attr='ti
         prepend_attr(str): whether another document attribute, such as the title of the document, to each passage, following [Dai2019]. Defaults to 'title'. 
         title_attr(str): what is the name of the dataframe attribute containing the title the document to be split into passages.
             Default is 'title'. Only used if prepend_title is set to True.
-        tokenizer(obj): which model to use for tokenizing. the object must have a `.tokenize(str) -> list[str]` method for tokenization.
+        tokenizer(obj): which model to use for tokenizing. The object must have a `.tokenize(str) -> list[str]` method for tokenization and `.convert_tokens_to_string(list[str]) -> str` for detokenization.
             Default is None. Tokenisation is perfomed by splitting on one-or-more spaces, i.e. based on the Python regular expression ``re.compile(r'\s+')``
     
     Example::
@@ -409,17 +409,8 @@ class SlidingWindowPassager(Transformer):
 
         # check if the tokenizer has the `.tokenize()` and the `.convert_tokens_to_string()` method
         if self.tokenizer is not None:
-            tokenizer_attr = getattr(self.tokenizer, 'tokenize', False)
-            if not tokenizer_attr or not callable(tokenizer_attr):
-                raise TypeError("%s doesn't have a `tokenize(str) -> list[str]` method. The tokenizer must have `tokenize(str) -> list[str]` method.")
-            else:
-                self.tokenize = self.tokenizer.tokenize
-            
-            convert_to_string_attr = getattr(self.tokenizer, 'convert_tokens_to_string', False)
-            if not convert_to_string_attr or not callable(convert_to_string_attr):
-                raise TypeError("%s doesn't have a `convert_tokens_to_string(list[str]) -> str` method. The tokenizer must have `convert_tokens_to_string(list[str]) -> str` method.")
-            else:
-                self.detokenize = self.tokenizer.convert_tokens_to_string
+            self.tokenize = self.tokenizer.tokenize
+            self.detokenize = self.tokenizer.convert_tokens_to_string
         else:
             self.tokenize = re.compile(r"\s+").split
             self.detokenize = ' '.join
--- a/pyterrier/text.py
+++ b/pyterrier/text.py
@@ -432,15 +432,11 @@ class SlidingWindowPassager(Transformer):
         return self.applyPassaging_no_qid(topics_and_res)
 
     def applyPassaging_no_qid(self, df):
-        if self.tokenizer is None:
-            p = re.compile(r"\s+")
+        tokenize = re.compile(r"\s+").split if self.tokenizer is None else self.tokenizer.tokenize
         rows=[]
         for row in df.itertuples():
             row = row._asdict()
-            if self.tokenizer is None:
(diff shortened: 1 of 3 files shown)


mihirs16:
yet to discuss the possibility and implementation of sentence segmentation.

cmacdonald:
Thanks @mihirs16 - we're attending ECIR this week, so reviewing of the PR may be delayed.

pyterrier/text.py
```diff
@@ -414,11 +432,15 @@ def transform(self, topics_and_res):
...
+        if self.tokenizer is None:
+            p = re.compile(r"\s+")
         rows=[]
         for row in df.itertuples():
             row = row._asdict()
-            toks = p.split(row[self.text_attr])
+            if self.tokenizer is None:
+                toks = p.split(row[self.text_attr])
+            else:
+                toks = self.tokenizer.tokenize(row[self.text_attr])
```
cmacdonald, line 443: why note assign a local tokenizer variable:

```python
tokenizer = p.split if self.tokenizer is None else self.tokenizer.tokenize 

for bla:
  toks = tokenizer(row[self.text_attr])
```

This take an if statement out of the critical loop

pyterrier/text.py
```diff
@@ -463,7 +486,10 @@ def applyPassaging(self, df, labels=True):
                     rank=0
                     currentQid = qid
                 rank+=1
-                toks = p.split(row[self.text_attr])
+                if self.tokenizer is None:
+                    toks = p.split(row[self.text_attr])
+                else:
+                    toks = self.tokenizer.tokenize(row[self.text_attr])
```
cmacdonald, line 492: as per above

cmacdonald (changes requested):
Can you add some unit new tests too? You should assert that your requirements on the token lengths are assured. 
you may need to add transformers to requirements-test.txt (but ideally, I'd like NOT to add such a heavy dependency), so perhaps you could mock up another tokenizer class.

pyterrier/text.py
mihirs16: i've fixed this and added a unit-test as well

mihirs16:
hey! would love further feedback (if any) about this

cmacdonald:
Hi @mihirs16 

Thanks for the ping. Most of the PR looks good. I think that my concern is whether 
```python
row[self.text_attr] = ' '.join(toks)
```
is now appropriate for recovering the passage segment.

Perhaps if the tokenizer is specified, it should call the .decode() method to recover the passage? [NB: Not sure .decode() is the correct method!)

I think also we could have a separate test case that runs IF transformers is installed, and is skipped otherwise:
```
def test_sliding_tokenize_HGF(self):
  try:
    from transformers import AutoTokenizer
   catch:
    self.skipTest("transformers not installed") 
  ...
```

mihirs16:
Sorry for the long post ahead 😅 but most of it is just python shell output.

> Thanks for the ping. Most of the PR looks good. I think that my concern is whether
> 
> ```python
> row[self.text_attr] = ' '.join(toks)
> ```
> 
> is now appropriate for recovering the passage segment.
> 
> Perhaps if the tokenizer is specified, it should call the .decode() method to recover the passage? [NB: Not sure .decode() is the correct method!)

@cmacdonald that's a very good point, here's a comparison between the first time a passage is tokenized, and then de-tokenized using `' '.join(toks)`, and then tokenized again to see if it reproduces the same result... which it does not.

```python
>>> tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
>>> tokenized = tokenizer.tokenize("Perhaps if the tokenizer is specified, it should call the .decode() method to recover the passage?")
>>> tokenized, len(tokenized)
(['perhaps', 'if', 'the', 'token', '##izer', 'is', 'specified', ',', 'it', 'should', 'call', 'the', '.', 'deco', '##de', '(', ')', 'method', 'to', 'recover', 'the', 'passage', '?'], 23)
>>> untokenized = ' '.join(tokenized)
>>> untokenized
'perhaps if the token ##izer is specified , it should call the . deco ##de ( ) method to recover the passage ?'
>>> tokenized_again = tokenizer.tokenize(untokenized)
>>> tokenized_again, len(tokenized_again)
(['perhaps', 'if', 'the', 'token', '#', '#', 'i', '##zer', 'is', 'specified', ',', 'it', 'should', 'call', 'the', '.', 'deco', '#', '#', 'de', '(', ')', 'method', 'to', 'recover', 'the', 'passage', '?'], 28)
```


As far as i understand, `.decode()` is useful when we have used `.encode()` to tokenized and encode the passage. Here's the following example using a huggingface transformer:

```python
>>> encoded = tokenizer.encode("Perhaps if the tokenizer is specified, it should call the .decode() method to recover the passage?")
>>> encoded, len(encoded)
([101, 3383, 2065, 1996, 19204, 17629, 2003, 9675, 1010, 2009, 2323, 2655, 1996, 1012, 21933, 3207, 1006, 1007, 4118, 2000, 8980, 1996, 6019, 1029, 102], 25)
>>> decoded = tokenizer.decode(encoded)
>>> decoded
'[CLS] perhaps if the tokenizer is specified, it should call the. decode ( ) method to recover the passage? [SEP]'
```

Otherwise, if we use the `.tokenize()` method, there are other utilities such as `.convert_tokens_to_string()` to get the original input again. Here's an example:

```python
>>> tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
>>> tokenized = tokenizer.tokenize("Perhaps if the tokenizer is specified, it should call the .decode() method to recover the passage?")
>>> tokenized, len(tokenized)
(['perhaps', 'if', 'the', 'token', '##izer', 'is', 'specified', ',', 'it', 'should', 'call', 'the', '.', 'deco', '##de', '(', ')', 'method', 'to', 'recover', 'the', 'passage', '?'], 23)
>>> untokenized = tokenizer.convert_tokens_to_string(tokenized)
>>> untokenized
'perhaps if the tokenizer is specified, it should call the. decode ( ) method to recover the passage?'
>>> tokenized_again = tokenizer.tokenize(untokenized)
>>> tokenized_again, len(tokenized_again)
(['perhaps', 'if', 'the', 'token', '##izer', 'is', 'specified', ',', 'it', 'should', 'call', 'the', '.', 'deco', '##de', '(', ')', 'method', 'to', 'recover', 'the', 'passage', '?'], 23)
```

In the light of this, I suppose the `Passager` class should have a `detokenize` to handle this in the same way as `tokenize`?

(using `.convert_tokens_to_string()` and making it a mandatory attribute, just like `tokenize()` for the `tokenizer` passed to `SlidingWindowPassager`).

cmacdonald:
I think the type of members of toks could be string or int.

So I think we can roughly do:
```python
toks = tok.tokenize(text)
reconstructed = tok.convert_tokens_to_string(toks)
```

mihirs16:
i took the liberty of setting tokenize and detokenize functions in the constructor itself, and also added a separate unittest to handle the case where transformers are installed (but haven't added it to the `requirements-test.txt`)

mihirs16:
@cmacdonald would love some feedback on this!

pyterrier/text.py
```diff
@@ -386,14 +397,32 @@ def __init__(self, **kwargs):
...
         self.passage_length = passage_length
         self.passage_stride= passage_stride
         self.join = join
         self.prepend_title = prepend_title
+        self.tokenizer = tokenizer
+
+        # check if the tokenizer has the `.tokenize()` and the `.convert_tokens_to_string()` method
+        if self.tokenizer is not None:
+            tokenizer_attr = getattr(self.tokenizer, 'tokenize', False)
+            if not tokenizer_attr or not callable(tokenizer_attr):
```
cmacdonald, line 413: not sure why being callable is relevant, given we expect it to have tokenize() and convert_tokens_to_string() are expected.

pyterrier/text.py
```diff
@@ -176,14 +177,23 @@ def sliding( text_attr='body', length=150, stride=75, join=' ', prepend_attr='ti
         prepend_attr(str): whether another document attribute, such as the title of the document, to each passage, following [Dai2019]. Defaults to 'title'. 
         title_attr(str): what is the name of the dataframe attribute containing the title the document to be split into passages.
             Default is 'title'. Only used if prepend_title is set to True.
+        tokenizer(obj): which model to use for tokenizing. the object must have a `.tokenize(str) -> list[str]` method for tokenization.
```
cmacdonald, line 180: We should be more specific about the requirements - it must have a convert_tokens_to_string() method too

cmacdonald (commented):
Sorry for the delay on this. I have left a few comments, but its nearly there :-)

pyterrier/text.py
mihirs16: as mentioned, `.tokenize()` and `.convert_tokens_to_string()` are expected, so both these checks are to enforce this requirement and make sure they are functions and throw an error if they aren't. we can remove it since it'll throw an error anyways when being called

cmacdonald:
Thanks for this @mihirs16!

Issue #393, opened 2023-08-30
Inprecise docstring for index class


lukaszett:
**Is your feature request related to a problem? Please describe.**
The docstring for the different index classes reads as follows with regards to metadata to record:
```
meta (Dict[str,int]) – What metadata for each document to record in the index, and what length to reserve. Defaults to {“docno” : 20, “filename” : 512}.
```
I feel like the word "reserve" is inprecise here as it does not make it clear enough that any inputs longer than the reserved length will be omitted. Both a [comment](https://github.com/terrier-org/pyterrier/blob/e47970a3a419e0580f02763b06c996f9c1ed5701/pyterrier/index.py#L171C1-L171C1) as well as the [terrier documentation](http://terrier.org/docs/v4.0/javadoc/org/terrier/indexing/TaggedDocument.html) are much more precise (terrier calls this "max_length" instead of reserved length.)

**Describe the solution you'd like**
The docstring should contain the information that metadata is only stored up to the defined length.

**Describe alternatives you've considered**
None

**Additional context**
None

seanmacavaney:
Thanks! I agree that this behaviour is worth clarifying in the documentation. I've merged the PR.

PR #394, opened 2023-08-30
clarify docstring for indexing with regards to metadata

lukaszett:
see #393


seanmacavaney (approved):
Thanks! Yeah, I agree that this behaviour be more clearly expressed.

I made a minor tweak (cropped -> truncated) to be slightly more precise.

PR #411, opened 2023-11-07
Add troubleshoot for installation and certification error

Krissy510:
I tried to install PyTerrier on macOS (M1) and Ubuntu 22.04 (Docker). However, both installations failed because `pyautocorpus` could not find `pcre`. After installing `pcre` I can install PyTerrier successfully.

After finished installing, for macOS there's an error that will cause due to a certification error. To solve this issue the user will need to install a certificate which is located in the Python folder.

These solutions and errors were not presented in the current documentation which is why I suggest adding a troubleshooting guide in case any users have the same issue.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -23,6 +23,21 @@ If you want the latest version of PyTerrier, you can install direct from the Git
 
 NB: There is no need to have a local installation of the Java component, Terrier. PyTerrier will download the latest release on startup.
 
+Installation Troubleshoot
+============
+
+If the installation failed due to `pyautocorpus` did not run successfully. You will need to install `pcre` to your machine.
+
+macOS::
+
+    brew install pcre
+
+Linux::
+
+    apt-get update -y
+    apt-get install libpcre3-dev -y
+
+
 Configuration
 ==============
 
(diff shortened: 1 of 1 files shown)


cmacdonald:
Thanks @Krissy510 - do you have a copy of the exceptions you saw.

https://github.com/seanmacavaney/pyautocorpus/issues/3 seems related @seanmacavaney

Krissy510:
> Thanks @Krissy510 - do you have a copy of the exceptions you saw.
> 
> [seanmacavaney/pyautocorpus#3](https://github.com/seanmacavaney/pyautocorpus/issues/3) seems related @seanmacavaney

Yes, I do. This example is what I got from trying to install PyTerrier on Ubuntu 22.04. In addition, this also happened on macOS.
error message

cmacdonald:
Yes, we dont have MacM1 builds for pyautocorpus (blame github :-( )
For ubuntu, I would have expected the build binaries to have worked, as there is a Python 3.10 manylinux build - see https://pypi.org/project/pyautocorpus/0.1.12/#files

Krissy510:
> Yes, we dont have MacM1 builds for pyautocorpus (blame github :-( ) For ubuntu, I would have expected the build binaries to have worked, as there is a Python 3.10 manylinux build - see https://pypi.org/project/pyautocorpus/0.1.12/#files

I understood that, but the point of this pull request is to add a section to give a guide to troubleshooting issues that are presented. After installing `pcre` I was able to install it successfully. In addition, for the certification error, I think it is a good idea to write this down in the installation guide for macOS users.

cmacdonald:
Yes, I understood that, and we appreciate the PR. We may merge, but at least for the pyautocorpus there may be an underlying problem.

cmacdonald:
@seanmacavaney pointed out that your arch is M1 even for Ubuntu, hence explaining why pyautocorpus needs compiled. We may drop pyautocorpus as a required dependency in ir_datasets, but thats for the future.

PR #414, opened 2023-11-09
fix parsing of trecxml topics

lukaszett:
The current implementation of `_read_topics_trecxml` is expecting the topic number to be supplied via attribute to the topic tag. However, this does not seem to be the current way TREC is formatting topics in xml. See for example the [health misinfo topics](https://trec.nist.gov/data/misinfo/misinfo-2021-topics.xml):
```
<topic>
<number>101</number>
<query>ankle brace achilles tendonitis</query>
<description>
Will wearing an ankle brace help heal achilles tendonitis?
</description>
<narrative>
Achilles tendonitis is a condition where one experiences pain in the Achilles tendon located near the heel. An ankle brace is usually worn around the ankles to protect and limit movement. A very useful document would discuss the effectiveness of using ankle braces to help heal Achilles tendonitis. A useful document would help a user make a decision about the use of ankle braces for treating tendonitis by providing information about recommended treatments for Achilles tendonitis, ankle braces, or both.
</narrative>
<disclaimer>
We do not claim to be providing medical advice, and medical decisions should never be made based on the stance we have chosen. Consult a medical doctor for professional advice.
</disclaimer>
<stance>unhelpful</stance>
<evidence>
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3134723/
</evidence>
</topic>
```

My change should not break backwards compatibility, also added a test to confirm this (see topic 3 still using an attribute to set the number).

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,3 +162,4 @@ By downloading and using PyTerrier, you agree to cite at the undernoted paper de
  - Sarawoot Kongyoung, University of Glasgow
  - Zhan Su, Copenhagen University
  - Marcus Schutte, TU Delft
+ - Lukas Zeit-Altpeter, Friedrich Schiller University Jena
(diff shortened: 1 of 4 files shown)


cmacdonald:
Nice job. Could you add your name and affiliation to the README.md list please?

> (see topic 3 still using an attribute to set the number).

Could you add an XML comment to the test file that says this. `<!-- Use attribute rather than tag -->` or something similar.

cmacdonald:
super, thanks @lukaszett

PR #420, opened 2023-12-18
paird ttost

seanmacavaney:
The unit test should probably be a paired ttest rather than an independent one, given the paired nature of queries.


cmacdonald:
Good spot!

PR #421, opened 2024-01-11
read_results optimization

seanmacavaney:
Per my testing, this speeds up read_results by around 50% for a reasonably large result file (`msmarco-passage/dev/small` results with 1000 per query).


cmacdonald:
lgtm! thanks

Issue #425, opened 2024-02-12
get_corpus_iter() from a Terrier index


cmacdonald:
pyterrier_pisa has a get_corpus_iter(), courtesy of @seanmacavaney 

It would be useful to have such a function for a Terrier index.

A few options that I can foresee:

1.  Terrier's Index class can expose a single get_corpus_iter() method that has pretokenized option. 
If pretokenized is true a version of the direct index.
If pretokenized is false, its a version of the meta index.

2. Expose these in the relevant classes:
a. direct index exposes a get_corpus_iter() which is pretokenized
b. meta index exposes  a get_corpus_iter() which is metadata

3. Both!

cmacdonald:
So option 3?

seanmacavaney:
I'd like the default to be to return everything that's available. If there's a case where that isn't needed, we can add options later.

cmacdonald:
#430 merged

PR #426, opened 2024-02-14
added corpus_iter for Terrier index

cmacdonald:
Initial draft of https://github.com/terrier-org/pyterrier/issues/425

Unit tests are needed. 

@seanmacavaney is the API useful?

### DIFF
--- a/docs/terrier-index-api.rst
+++ b/docs/terrier-index-api.rst
@@ -171,3 +171,31 @@ of the term (obtained from the Lexicon, in the form of the LexiconEntry), as wel
         print("%s with score %0.4f"  % (docno, score))
 
 Note that using BatchRetrieve or similar is probably an easier prospect for such a use case.
+
+Can I get the index as a corpus_iter()?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A corpus_iter can be obtained from an Index object, which allows for instance:
+ - indexing the pre-tokenised Terrier index directly in another indexing pipeline
+ - extracting document metadata for ingestion into another indexing pipeline
+
+Metadata Example::
+
+    iter = index.get_corpus_iter(direct=False)
+    next(iter)
+    ## could display {'docno' : 'd1', 'text' : 'This document contains ...' }
+
+Pre-tokenised Example::
+
+    iter = index.get_corpus_iter()
+    next(iter)
+    ## could display {'docno' : 'd1', 'toks' : {'a' : 1, 'the' : 2}}
+
+Document Pruning Example::
+
+    index_pipe = (
+        # make a new toks column for each document, keeping only terms with frequency > 1
+        pt.apply.toks(lambda row: { t : row['toks'][t] for t in row['toks'] if row['toks'][t] > 1 } ) 
+        >> pt.IterDictIndexer("./pruned_index", pretokenised=True)
+    )
+    new_index_ref = index_pipe.index( index.get_corpus_iter())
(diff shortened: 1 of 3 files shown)


pyterrier/bootstrap.py
```diff
@@ -271,14 +271,55 @@ def _index_add(self, other):
...
+            keys_offset = {'docno' : keys_offset['docno'] }
+            direct_inputstream = self.getIndexStructureInputStream("direct")
+            lex = self.getLexicon()
+
+            while direct_inputstream.hasNext():
+                ip = direct_inputstream.getNextPostings() #this is the next() method
+
+                for skipped in range(0, direct_inputstream.getEntriesSkipped()):
+                    meta = meta_inputstream.next()
+                    yield {k : meta[keys_offset[k]] for k in keys_offset}
```
seanmacavaney, line 297: Not super familar with the semantics of `getEntriesSkipped`. But should this include an empty `toks` to ensure that every dict from the iter has the same fields?

pyterrier/bootstrap.py
```diff
@@ -271,14 +271,55 @@ def _index_add(self, other):
             raise ValueError("Cannot document-wise merge indices with and without positions (%r vs %r)" % (blocks_1, blocks_2))
         multiindex_cls = autoclass("org.terrier.realtime.multi.MultiIndex")
         return multiindex_cls([self, other], blocks_1, fields_1 > 0)
+    
+    def _index_corpusiter(self, direct=True):
```
seanmacavaney, line 275: `direct` isn't the most intuitive name to me for this field. Perhaps `return_toks`?

docs/terrier-index-api.rst
```diff
@@ -171,3 +171,31 @@ of the term (obtained from the Lexicon, in the form of the LexiconEntry), as wel
...
+
+    iter = index.get_corpus_iter(direct=False)
+    next(iter)
+    ## could display {'docno' : 'd1', 'text' : 'This document contains ...' }
+
+Pre-tokenised Example::
+
+    iter = index.get_corpus_iter()
+    next(iter)
+    ## could display {'docno' : 'd1', 'toks' : {'a' : 1, 'the' : 2}}
```
seanmacavaney, line 192: should also include "text"?

seanmacavaney (commented):
looks good, just a few minor suggestions

pyterrier/bootstrap.py
cmacdonald: it should, yes

docs/terrier-index-api.rst
cmacdonald: if text is present in the meta index

pyterrier/bootstrap.py
cmacdonald: or `toks`? The corresponding indexing arg is `pretokenised`?

pyterrier/bootstrap.py
cmacdonald: > Not super familar with the semantics of getEntriesSkipped.

direct_inputstream only address documents that are non-empty. So we have to address non-empty documents using getEntriesSkipped()

cmacdonald:
> direct isn't the most intuitive name to me for this field.

Final decision: `toks` or `return_toks` ?

seanmacavaney:
I don't think I feel super strongly one way or the other, but I feel `return_toks` is a bit clearer (i.e., it sounds like a boolean).

cmacdonald:
Ok, ta. I write some unit tests then I merge.

cmacdonald:
My revised implementation requires Python 3.8.
Python 3.7 is EOL (July 2023). Should enforce an upgrade?

seanmacavaney:
I'm happy with min python version of 3.8

PR #441, opened 2024-05-09
Add RankVicuna and RankZephyr Plugins

kaustubhdhole:


seanmacavaney:
Awesome, thanks @kaustubhdhole!

A quick tip: If you change `LLMReRanker` to inherit from `pt.Transformer` and rename `rerank_pyt` to `transform`, you can avoid the lambda. And if you use `pt.text.get_text()` in the pipeline, it'll add the document's text into the dataframe, so you don't need to load them all into memory.

kaustubhdhole:
Thanks @seanmacavaney for merging and for the suggestion! I will update the other repo.

PR #447, opened 2024-07-20
pt.java

seanmacavaney:
WIP. Draft pull request for now (to get GH action testing).


cmacdonald:
Does jnius remain a declared dependency?

seanmacavaney:
> Does jnius remain a declared dependency?

Long-term, I hope that we can make it optional.

In the short term, it feels too ingrained and is effectively required to do nearly anything. (At least this is how it's felt as I've worked to integrate this PR.)

So for this PR, I'd say let's try to aim for: "pt.init() not required, but jnius is"

pyterrier/java.py
```diff
@@ -0,0 +1,171 @@
...
+def _legacy_post_init(jnius):
+    pt.HOME_DIR = pt.io.pyterrier_home()
+    pt.properties = autoclass('java.util.Properties')()
+    pt.ApplicationSetup = autoclass('org.terrier.utility.ApplicationSetup')
+    pt.autoclass = jnius.autoclass
+    pt.cast = jnius.cast
+
+
+@pt.utils.once()
+def init() -> None:
```
cmacdonald, line 21: should this not be private?

seanmacavaney: I'd like a way to force init if you want to. E.g., you might want to be sure that java has started before timing stuff.

seanmacavaney:
I made a nice enhancement along the way: now if the download fails, `mavenresolver` uses the most recent version found locally. A warning is emitted in this case. The behavior can be overridden by setting `mavenresolver.online()`, which will raise an error in this case instead.

seanmacavaney:
Note to self: I still need to update the documentation/docstrings/etc.

seanmacavaney:
Okay, I can't figure out what the issue is with the `macos-latest` test runner errors. I've tested on Intel, M1, and M3 macs without issue, matching the OS version and all.

seanmacavaney:
Alright, I think this is ready for review @cmacdonald. It turned out to be a bit more involved than I was expecting, but I feel the changes are important.

Main changes:
 - `pt.init()` no longer required to be called. This is accomplished by marking up functions/classes that need java with a new `@pt.java.required` decorator. This decorator performs a `if not started: init()` check automatically.
 - How is the configuration that used to be done via the `pt.init()` arguments handled? There are new functions under the `pt.java` that allow for these to be set, e.g., `pt.java.set_memory_limit()` or `pt.java.add_jar()`.
 - Split out Terrier stuff into `pt.terrier`, and Anserini stuff into `pt.anserini`. This sets the stage for potentially splitting out these packages entirely, which will mean we could then make Java optional on the core pyterrier package.
 - This version should be backward-compatible: `pt.init()` still exists (but is deprecated) and sets the corresponding settings such as `pt.java.set_memory_limit()`. Similarly, functions like `pt.autoclass` still exist in the main namespace and are deprecated in favor of `pt.java` alternatives.

Extras/notes/etc:
 - Automated build checks to ensure that `@pt.java.required` is applied everywhere it should be.
 - You no longer need to set `pt.init(no_download=True)` when working offline to avoid failures with `mavenresolver`; it will fall back on the most recent cached version if it detects a download failure (emitting a warning). If it can't find one, it raises an error. This offline behaviour can be disabled to force it into online mode, just in case if that's ever desired.
 - The anserini version pulls its jar from the version distributed with pyserini by default, allowing us to support more versions without compatibility problems. It also now tests the latest version too.
 - Substantially cleaned up `__init__.py` by moving almost everything to subpackages.
 - `pt.Utils` now links to `pt.utils` (I had to add some utilities for this and I didn't want to put them under the deprecated namespace)
 - `pt.io.pyterrier_home()` to replace `pt.HOME_DIR`
 - Most packages now `import pyterrier as pt` at the top level, eliminating the need for lots of specific imports, inline function imports, etc.
 - Parallel mode should be easier now with `pt.java.parallel_init` and `pt.java.parallel_init_args()` functions that can plug in directly to corresponding arguments in e.g., `multiprocesisng.Pool`.
 - New `JavaClasses` object to avoid long initlizers with loads of `pt.autoclass` calls.

Caveats:
 - I've spent loads of time trying to get the parallel tests passing consistently on GitHub Actions. Occasionally they fail, and they always seem to fail on `macos-latest`. However, I've never been able to reproduce the issue in any environment outside GitHub Actions, including on the latest macos on both an M1 and M3 chips. So I think it's something weird about the runners, rather than a problem with the code. It's considered an ["experimental feature"](https://pyterrier.readthedocs.io/en/latest/parallel.html) anyway.
 - This is clearly a big change. If merged, let's not rush to push to pypi and instead use this version locally for a bit to ensure nothing unexpected crops up in extension packages, etc.

PR #462, opened 2024-08-16
Parallel fixes

seanmacavaney:
Fixes bug where pyterrier would re-download the terrier jar when in snapshot mode and running in parallel


cmacdonald (approved):
lgtm

PR #465, opened 2024-08-20
No jinja

cmacdonald:
Trial removing jinja2 dependency (what is this used for?)


cmacdonald:
This PR adds comments to requirements.txt so we know what each dependency does.

I made appropriate changes in setup.py to adequately parse said requirements.txt 

Comments are allowed in requirements.txt - see https://stackoverflow.com/a/9160112

seanmacavaney (approved):

PR #476, opened 2024-08-23
deprecated modules

seanmacavaney:
to handle existing usage of the following without breaking

```python
import pyterrier.batchretrieve
from pyterrier.batchretrieve import X
import pyterrier.index
from pyterrier.index import X
```

The functions/classes included in each module were mined using a search over public GitHub, plus anything else that seemed likely to be used.

### DIFF
--- a/pyterrier/batchretrieve.py
+++ b/pyterrier/batchretrieve.py
@@ -1,0 +1,25 @@
+"""DEPRECATED MODULE
+
+Warning: This module is deprecated and will be removed in a future version.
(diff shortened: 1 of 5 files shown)


cmacdonald:
lgtm. have you checked that the deprecated warnings appear as expected?

seanmacavaney:
Everything except for `IndexingType.X` (unless you call a method on one). I expect `IndexingType` will be sorted when moving the indexer.

PR #480, opened 2024-09-04
Transformer.compile improvements

### LINKED ISSUE #451: Transformer.compile improvements
**Current Approach**

Currently, we use `matchpy` to support pipeline optimization. A set of replacement rules are defined in `pyterrier.transformer.rewrite_rules`, which all get executed when `Transformer.compile()` is called.

`Transformer.compile` is not used heavily, though it can be important for getting the most out of pipelines.

**Problems with the current approach**

The current approach has several downsides:
 - It decouples the logic for applying combination rules from the transformers themselves. I.e., I'd expect a `ComposedPipeline` to define how it can be combined/optimized, not some other place in the code.
 - It requires maintaining the global `rewrite_rules` variable.
 - It introduces additional complexity into the class hierarchy, with certain transformers defined as `matchpy.Operations` with an `arity` attribute.
 - Overall, `matchpy` feels like overkill for our needs, which are relatively simple combination rules. Using it introduces a dependency that we probably don't need.

**Proposed Solution**

I propose dropping `matchpy` in favor of defining optimization logic in individual transformers. This will be faciliated with three methods:

 - `Transformer.compile(self) -> Transformer` -- existing method, but now individual transformers are responsible for implementing it. By default, it returns the transformer itself, but can be overridden with other functionality, such as merging `ComposedPipeline` together.
 - `Transformer.fuse_left(self, left: Transformer) -> Optinal[Transformer]` -- Fuses this transformer with a transformer that immediately precedes this one in a pipeline. The new transformer should have the same effect as performing the two transformers in sequence, i.e., `pipeline_unfused` and `pipeline_fused` in the following example should provide the same results:
```python
pipeline_unfused = left >> self
pipeline_fused = self.fuse_left(left)
```
 - `Transformer.fuse_right(self, right: Transformer) -> Optinal[Transformer]` -- Similar to `fuse_left` but for a transformer that immediately follows this one.

I'd expect that most pipelines wouldn't need to implement any of these methods. Perhaps only `ComposedPipeline` would define `compile`, which would be responsible for performing the merging logic, something like:

```python
def compile(self) -> Transformer:
    out = deque()
    inp = deque(tuple(chain.from_iterable((t.models if isinstance(t, ComposedPipeline) else (t,)) for t in self.models)))
    while inp:
        right = inp.popleft()
        if out and (fused_right := out[-1].fuse_right(right)):
            out.pop()
            inp.appendleft(fused_right)
        elif out and (fused_left := right.fuse_left(out[-1])):
            out.pop()
            inp.appendleft(fused_left)
        else:
            out.append(right)
    return ComposedPipeline(*out)
```

An example implemetnation of `fuse_left` that merges a `BatchRetrieve` into a `FeaturesBatchRetrieve`, using the current logic:

```python
class FeaturesBatchRetrieve:
  ...
  def fuse_left(self, left: pt.Transformer) -> pt.Transformer:
      if isinstance(left, BatchRetrieve) and if self.indexref == left.indexref:
          controls = dict(controls)
          controls['wmodel'] = left.wmodel
          return FeaturesBatchRetrieve(
              self.indexref,
              self.features,
              controls,
              self.properties,
              self.threads,
          )
```

seanmacavaney:
wip

fixes  #451

### DIFF
--- a/pyterrier/ops.py
+++ b/pyterrier/ops.py
@@ -287,6 +287,26 @@ class FeatureUnionPipeline(NAryTransformerBase):
         assert not "features_y" in final_DF.columns 
         return final_DF
 
+    def compile(self) -> Transformer:
+        """
+            Returns a new transformer that fuses feature unions where possible.
+        """
+        out = deque()
+        inp = deque([t.compile() for t in self._transformers])
+        while inp:
+            right = inp.popleft()
+            if out and isinstance(out[-1], SupportsFuseFeatureUnion) and (fused := out[-1].fuse_feature_union(right, is_left=True)) is not None:
+                out.pop()
+                inp.appendleft(fused)
+            elif out and isinstance(right, SupportsFuseFeatureUnion) and (fused := right.fuse_feature_union(out[-1], is_left=False)) is not None:
+                out.pop()
+                inp.appendleft(fused)
+            else:
+                out.append(right)
+        if len(out) == 1:
+            return out[0]
+        return FeatureUnionPipeline(*out)
+
     def __repr__(self):
         return '(' + ' ** '.join([str(t) for t in self._transformers]) + ')'
 
@@ -353,15 +373,15 @@ class ComposedPipeline(NAryTransformerBase):
             Returns a new transformer that fuses adjacent transformers where possible.
         """
         out = deque()
-        inp = deque(self._transformers)
+        inp = deque([t.compile() for t in self._transformers])
         while inp:
             right = inp.popleft()
-            if out and isinstance(out[-1], SupportsFuseRight) and (fused_right := out[-1].fuse_right(right)) is not None:
+            if out and isinstance(out[-1], SupportsFuseRight) and (fused := out[-1].fuse_right(right)) is not None:
                 out.pop()
-                inp.appendleft(fused_right)
-            elif out and isinstance(right, SupportsFuseLeft) and (fused_left := right.fuse_left(out[-1])) is not None:
+                inp.appendleft(fused)
+            elif out and isinstance(right, SupportsFuseLeft) and (fused := right.fuse_left(out[-1])) is not None:
                 out.pop()
-                inp.appendleft(fused_left)
+                inp.appendleft(fused)
             else:
                 out.append(right)
         if len(out) == 1:
@@ -426,4 +446,16 @@ class SupportsFuseRankCutoff(Protocol):
         relaxed. In this case, it is preferred to return `self`.
 
         If the fusion is not possible, `None` should be returned.
+        """
+
+
+@runtime_checkable
+class SupportsFuseFeatureUnion(Protocol):
+    def fuse_feature_union(self, other: 'Transformer', is_left: bool) -> Optional['Transformer']:
+        """Fuses this transformer with another one that provides features.
+
+        This method should return a new transformer that is equivalent to performing self ** other, or `None`
+        if the fusion is not possible.
+
+        is_left is True if self's features are to the left of other's. Otherwise, self's features are to the right.
         """
--- a/pyterrier/ops.py
+++ b/pyterrier/ops.py
@@ -1,46 +1,38 @@
-from .transformer import Transformer, Estimator, get_transformer, Scalar
+from .transformer import Transformer, Estimator, get_transformer
 from .model import add_ranks
-from matchpy import Operation, Arity
+from collections import deque
 from warnings import warn
-from typing import Iterable
+from typing import Iterable, List, Optional, Protocol, runtime_checkable
+from itertools import chain
 import pandas as pd
 
-class BinaryTransformerBase(Transformer,Operation):
-    """
-        A base class for all operator transformers that can combine the input of exactly 2 transformers. 
-    """
-    arity = Arity.binary
-
-    def __init__(self, operands, **kwargs):
-        assert 2 == len(operands)
-        super().__init__(operands=operands,  **kwargs)
-        self.left = operands[0]
-        self.right = operands[1]
-
-class NAryTransformerBase(Transformer,Operation):
+class NAryTransformerBase(Transformer):
     """
         A base class for all operator transformers that can combine the input of 2 or more transformers. 
     """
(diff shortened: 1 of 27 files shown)


cmacdonald:
The following is an extract of a conversation between Sean and Craig

> what about transformer reuse?
> `br >> ( (a >> b) + (a >> c) )`
> could be rewritten as 
> `br >> a >> (b+c)`

That seems like an easy rule to add to CombSumTransformer's transformer (as long as we first convince ourselves that it's universally applicable):
```python
class CombSumTransformer:
  def compile(self):
    if isinstance(self.left, ComposedPipeline) and len(self.left) == 2 and isinstance(self.right, ComposedPipeline) and len(self.right) == 2 and self.left[0] == self.right[0]:
      return self.left[0] >> (self.left[1] + self.right[1])
```

In fact, it could be pretty easily generalized to any length prefix, which is impossible to express with matchpy afaik. Ie

`br >> ((a>>b>>c)+(a>>b>>d))`
->
`br >> a >> b >> (c+d)`

```python
class CombSumTransformer:
  def compile(self):
    if isinstance(self.left, ComposedPipeline) and isinstance(self.right, ComposedPipeline) and len(self.left) == len(self.right):
      prefix = takewhile(zip(self.left, self.right), lambda x: x[0] == x[1])
      if len(prefix) == len(self.left) - 1:
        return ComposedPipeline(prefix) >> (self.left[-1], self.right[-1])
```

pyterrier/transformer.py
```diff
@@ -260,7 +250,7 @@ def __and__(self, right : 'Transformer') -> 'Transformer':
     def __mod__(self, right : int) -> 'Transformer':
         assert isinstance(right, int)
         from .ops import RankCutoffTransformer
-        return RankCutoffTransformer(self, right)
+        return self >> RankCutoffTransformer(right)
```
cmacdonald, line 253: this changes the implementation of cutoff, and would change how it was rendered graphically. It also adds a level of indirection?

pyterrier/terrier/retriever.py
```diff
@@ -627,6 +640,8 @@ def __init__(self, index_location, features, controls=None, properties=None, thr
             raise ValueError("Multi-threaded retrieval not yet supported by FeaturesRetriever")
         
         super().__init__(index_location, controls, properties, **kwargs)
+        if self.wmodel is None and 'wmodel' in self.controls:
```
cmacdonald, line 643: whats the motivation for this change?

pyterrier/ops.py
```diff
@@ -1,46 +1,38 @@
...
 from .model import add_ranks
-from matchpy import Operation, Arity
+from collections import deque
 from warnings import warn
-from typing import Iterable
+from typing import Iterable, List, Optional, Protocol, runtime_checkable
+from itertools import chain
 import pandas as pd
 
-class BinaryTransformerBase(Transformer,Operation):
```
cmacdonald, line 8: have we lost something about Arity? is it not a useful concept to keep around? Is there anything that all operators inherit from?

pyterrier/transformer.py
seanmacavaney: This change adds flexibility to the `RankCutoffTransformer`. Now you can run it over an arbitrary dataframe directly, rather than needing a dummy transformer to run first. E.g.,:

```python
my_frame = pd.DataFrame(...)
# I want only the top 10 results of my_frame. Previously I'd have to:
my_frame_10 = RankCutoffTransformer(pt.Transformer.identity(), 10)(my_frame)
# Now I can:
my_frame_10 = RankCutoffTransformer(10)(my_frame)
```

I'd argue that this makes it better aligned with how most components in PyTerrier work—take input, transform, output.

Can you elaborate a bit on what you mean by indirection? I sorta see it as more direct to treat it as a regular pipeline component, rather than attach it to an (sometimes arbitrary) other transformer. I.e., defining the retrieval procedure as:

 1. Retrive
 2. Re-Rank
 3. Print results (debug)
 4. Take the top 10

Makes more sense to me than a procedure of:

 1. Retrive
 2. Re-Rank
 3. Print results (debug) AND take the top 10, consisting of:
     1. Print results (debug)
     2. Take the top 10

pyterrier/ops.py
seanmacavaney: I don't see operators as a subclass of transformers. To me, an operator is just syntactic sugar -- they make it easier to combine common transformers.

If the distinction is that an operator is a type of transformers that take other transformers as input, we do not consistently apply a base class as of now. E.g., some of the `pyterrier-caching` transformers take other transformers, as does `GAR`. Probably others too.

cmacdonald:
While we await the return of Pantar, is there any blockers preventing merging from your side Sean?

cmacdonald:
> I think ComposedPipeline should first call compile on each consistent transformer.

Makes sense

> Do we need the suffix Transformer in some of these operations? Cutoff(k) would look nicer rendered in a pipeline thanCutoffTransformer(k).

Agreed

> These would allow for the following optimization:

Happy to see more improvements like this.

cmacdonald:
- I wonder if ops.py should be renamed _ops.py? The operator transformer implementation arent really expected to be public
 - The protocols should be public though - i would think they should be in transformer.py and exposed in __init__.py

cmacdonald:
> I think ComposedPipeline should first call compile on each consistent transformer.

Doesnt this already happen? line 296 of _ops.py

https://github.com/terrier-org/pyterrier/pull/480/files#diff-f0762b1f9740e8c01dc0fad99d79297791cdd9d899fd61903a2c910e16c0b617R296

seanmacavaney:
Right, what I was thinking is that it needs to expand them out in the case where compile gives another ComposedPipeline. Otherwise the example won't work.

I think something like this should do the trick (but I haven't tested it):

```python
inp = deque(Compose(*(t.compile() for t in self._transformers))._transformers)
```

cmacdonald:
I committed my WIP to
https://github.com/terrier-org/pyterrier/compare/advanced_compile?expand=1

cmacdonald:
Nice job on this Sean!

PR #481, opened 2024-09-04
API change: Transformer.transform_iter() returns an iter-dict generator

cmacdonald:
# Problem

Currently `.transform_iter()` returns a DataFrame.

As a results, ComposedPipeline`.index()` calls `.transform_iter()` and has to undo the DataFrame for the next invocation.

The disadvantages are:
 - Extra DataFrame constructions and deconstructions
 - No API that users can use and implement that operates entirely in iter-dicts, and therefore avoid DataFrames

# Proposal

1. `.transform_iter()` takes and returns an iterable of dicts (which can be a generator), rather than a DataFrame
2. As a consequence of (1), `.__call__()` takes and returns an iter-dict (instantiated as a list), OR takes and returns a DataFrame. 
3. Implementers need only implement `.transform()` OR `.transform_iter()`. `.__call__()` can detect which one is implemented, redirect appropriately. 
4. Can we do the same thing for `.transform()` - default implementation is to call `.transform_iter()` (and vice versa)
5. Add an easy to remember kwarg for pt.apply to take iter-dicts (iter=True).

# Prerequisites

- [x] We need to be able to identify overridden methods; this needs to probably be done in the `Transformer.__init__()` constructor, rather than once per-invocation. https://stackoverflow.com/questions/9436681/how-to-detect-method-overloading-in-subclasses-in-python looks like a possible solution.

# Potential for breaking changes

We only told people to implement `.transform()` in the past, not `.transform_iter()`

However, I have found some repositories implementing `.transform_iter()`. See https://github.com/search?q=pyterrier+transform_iter&type=code, specifically
https://github.com/argonism/OPEX

Is this change enough to make this a major version change?

### DIFF
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -180,14 +180,21 @@ def generic(fn : Union[Callable[[pd.DataFrame], pd.DataFrame], Callable[[Iterabl
             fn(Callable): the function to apply to each row
             batch_size(int or None): whether to apply fn on batches of rows or all that are received
             verbose(bool): Whether to display a progress bar over batches (only used if batch_size is set, and iter is not set).
-            iter(bool): Whether to use transform_iter()
+            iter(bool): Whether to use the iter-dict API - if-so, then ``fn`` receives an iterable, and must return an iterator. 
 
-        Example::
-
-            # this transformer will remove all documents at rank greater than 2.
+        Example (dataframe)::
 
             # this pipeline would remove all but the first two documents from a result set
             pipe = pt.terrier.Retriever(index) >> pt.apply.generic(lambda res : res[res["rank"] < 2])
+
+         Example (iter-dict)::
+
+            # this pipeline would simlarly remove all but the first two documents from a result set
+            def _fn(iterdict):
+                for result in iterdict:
+                    if result["rank"] < 2:
+                        yield result
+            pipe = pt.terrier.Retriever(index) >> pt.apply.generic(_fn)
 
     """
     if iter:
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -243,7 +243,7 @@ def generic_apply(name, *args, drop=False, **kwargs) -> pt.Transformer:
         raise ValueError("Must specify a fn, e.g. a lambda")
 
     fn = args[0]
-    args=[]
+    args=()
 
     def _new_column(df):
         df[name] = df.apply(fn, axis=1, result_type='reduce')
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Callable, Any, Dict, Union, Iterator, Iterable, Sequence
+from typing import Callable, Any, Dict, Union, Iterable, Sequence
 import numpy.typing as npt
 import pandas as pd
 import pyterrier as pt
@@ -132,7 +132,7 @@ def doc_features(fn : Callable[[Union[pd.Series,Dict[str,Any]]], npt.NDArray[Any
     """
     return ApplyDocFeatureTransformer(fn, *args, **kwargs)
 
-def indexer(fn : Callable[[Iterator[Dict[str,Any]]], Any], **kwargs) -> pt.Indexer:
+def indexer(fn : Callable[[pt.model.IterDict], Any], **kwargs) -> pt.Indexer:
     """
         Create an instance of pt.Indexer using a function that takes as input an interable dictionary.
 
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -198,6 +198,8 @@ def generic(fn : Union[Callable[[pd.DataFrame], pd.DataFrame], Callable[[Iterabl
 
     """
     if iter:
+        if kwargs.get("add_ranks", False):
+            raise ValueError("add_ranks=True not supported with iter=True")
         return ApplyGenericIterTransformer(fn, *args, batch_size=batch_size, **kwargs)
     return ApplyGenericTransformer(fn, *args, batch_size=batch_size, **kwargs)
 
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -17,15 +17,17 @@ def _bind(instance, func, as_name=None):
     setattr(instance, as_name, bound_method)
     return bound_method
 
-def query(fn : Callable[[pd.Series], str], *args, **kwargs) -> pt.Transformer:
+def query(fn : Callable[[Union[pd.Series,Dict[str,Any]]], str], *args, **kwargs) -> pt.Transformer:
     """
         Create a transformer that takes as input a query, and applies a supplied function to compute a new query formulation.
 
         The supplied function is called once for each query, and must return a string containing the new query formulation.
-        Each time it is called, the function is supplied with a Panda Series representing the attributes of the query.
+        Each time it is called, the function is supplied with a Panda Series or dict representing the attributes of the query.
+        The particular type of the input is not controlled by the implementor, so the function should be written to support 
+        both, e.g. using ``row["key"]`` notation and not the ``row.key`` that is supported by a Series.
 
         The previous query formulation is saved in the "query_0" column. If a later pipeline stage is intended to resort to
-        be executed on the previous query formulation, a `pt.rewrite.reset()` transformer can be applied.  
+        be executed on the previous query formulation, a ``pt.rewrite.reset()`` transformer can be applied.  
 
         Arguments:
             fn(Callable): the function to apply to each row. It must return a string containing the new query formulation.
@@ -51,7 +53,7 @@ def query(fn : Callable[[pd.Series], str], *args, **kwargs) -> pt.Transformer:
                 ) >> pt.terrier.Retriever(index, wmodel="DPH")
 
         In both of the example pipelines above (`p1` and `p2`), the exact topics are not known until the pipeline is invoked, e.g.
-        by using `p1.transform(topics)` on a topics dataframe, or within a `pt.Experiment()`. When the pipeline 
+        by using `p1.transform(topics)` on a topics dataframe, or within a ``pt.Experiment()``. When the pipeline 
         is invoked, the specified function (`_remove_stops` in the case of `p1`) is called for **each** row of the 
         input datatrame (becoming the `q` function argument).
             
@@ -59,7 +61,7 @@ def query(fn : Callable[[pd.Series], str], *args, **kwargs) -> pt.Transformer:
     """
     return ApplyQueryTransformer(fn, *args, **kwargs)
 
-def doc_score(fn : Union[Callable[[pd.Series], float], Callable[[pd.DataFrame], Sequence[float]]], *args, batch_size=None, **kwargs) -> pt.Transformer:
+def doc_score(fn : Union[Callable[[Union[pd.Series,Dict[str,Any]]], float], Callable[[pd.DataFrame], Sequence[float]]], *args, batch_size=None, **kwargs) -> pt.Transformer:
     """
         Create a transformer that takes as input a ranked documents dataframe, and applies a supplied function to compute a new score.
         Ranks are automatically computed. doc_score() can operate row-wise, or batch-wise, depending on whether batch_size is set.
@@ -76,7 +78,7 @@ def doc_score(fn : Union[Callable[[pd.Series], float], Callable[[pd.DataFrame], 
 
             # this transformer will subtract 5 from the score of each document
             p = pt.terrier.Retriever(index, wmodel="DPH") >> 
-                pt.apply.doc_score(lambda doc : doc["score"] -5)
+                pt.apply.doc_score(lambda doc : doc["score"] -5) # doc["score"] works for both a dict and Series
 
         Can be used in batch-wise manner, which is particularly useful for appling neural models. In this case,
         the scoring function receives a dataframe, rather than a single row::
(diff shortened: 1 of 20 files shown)


pyterrier/transformer.py
```diff
@@ -54,8 +54,17 @@ class Transformer:
     """
         Base class for all transformers. Implements the various operators ``>>`` ``+`` ``*`` ``|`` ``&`` 
         as well as ``search()`` for executing a single query and ``compile()`` for rewriting complex pipelines into more simples ones.
+
+        Its expected that either ``.transform()`` or ``.transform_iter()`` be implemented by any class extending this - this rule
+        does not apply for indexers, which instead implement ``.index()``.
     """
 
+    def __init__(self, *args, **kwargs):
```
seanmacavaney, line 62: I worry about adding a constructor since there wasn't one previously, and people probably haven't been consistent in calling it. (I know I haven't.)

pyterrier/transformer.py
```diff
@@ -54,8 +54,17 @@ class Transformer:
...
+        super().__init__(*args, **kwargs) 
+        self._transform_implemented = type(self).transform != Transformer.transform
+        self._transform_iter_implemented = type(self).transform_iter != Transformer.transform_iter
+        # we cant test for either self._transform_implemented or  self._transform_iter_implemented here, due to indexers
```
seanmacavaney, line 66: Could this be checked when the subclass is defined using `__init_subclass__`?

I think indexers should probably define the methods in their base class to throw an error.

cmacdonald:
I didnt know about `__init_subclass__`, and on reading it sounds attractive, so I followed your suggestions:
```python
class Transformer:
    def __init_subclass__(cls, **kwargs):
        if cls.transform == Transformer.transform and cls.transform_iter == Transformer.transform_iter:
            raise NotImplementedError("You need to implement either .transform() or .transform_iter() in %s" % str(cls))
```

and

```python
class Indexer(Transformer):
    def index(self, iter : Iterable[dict], **kwargs):
         pass

    def transform(self, df):
        raise TypeError("Its not expected that .transform() is called on an Indexer")
        # TODO - what about classes that are both Indexers and Transformers?

    def transform_iter(self, df):
        raise TypeError("Its not expected that .transform_iter() is called on an Indexer")
        # TODO - what about classes that are both Indexers and Transformers?
```

This has two problems:
(1) pt.TransformerBase throws the error (as will any other "abstract" classes)
(2) We've now overridden `transform` and `transform_iter` in our Indexer, which would be a problem for a class that is both an Indexer and a Transformer? 

My code is in https://github.com/terrier-org/pyterrier/tree/transform_iter_subclass

seanmacavaney:
For (1) [`ABC`](https://docs.python.org/3/library/abc.html#abc.ABC) handles this somehow. Maybe we dig in a bit to see how it's done there?

(2) isn't so common, I suppose it could just be up to the implementer to make sure this is handled?

Or, as a more major change, maybe it doesn't make very much sense to have indexer inherit from transformer? Some of the transformer functionality applies, but actually not that much of it. They should be able to be included in a `ComposedPipeline`, but only at the end. Rank cutoff, feature composition, etc. do not apply. Etc.

cmacdonald:
> For (1) [ABC](https://docs.python.org/3/library/abc.html#abc.ABC) handles this somehow. Maybe we dig in a bit to see how it's done there?

They override `__new__`, which is much less common to override.

> (2) isn't so common, I suppose it could just be up to the implementer to make sure this is handled?

I've excluded Indexer from the check. However, it wont fail gracefully. 

> (3) Or, as a more major change, maybe it doesn't make very much sense to have indexer inherit from transformer? Some of the transformer functionality applies, but actually not that much of it. They should be able to be included in a ComposedPipeline, but only at the end. Rank cutoff, feature composition, etc. do not apply. Etc.

I'm amenable to exploring this, but not sure its relevant to this PR

cmacdonald:
I can confirm that the following works - apply `__call__` with an iterdict and it will return an iterdict 😀 :
```python
>>> r = pt.terrier.Retriever.from_dataset('vaswani', 'terrier_stemmed')
>>> (r % 1) ([{'qid' : 'q1', 'query' : 'chemical reactions'}])
[{'qid': 'q1', 'docid': 9373, 'docno': '9374', 'rank': 0, 'score': 10.619215559726808, 'query': 'chemical reactions'}]
```

pyterrier/apply_base.py
```diff
@@ -68,6 +69,39 @@ def transform(self, res):
...
+        assert not self.add_ranks
+        input = more_itertools.peekable(input)
+        try:
+            thisqid = input.peek()["qid"]
+        except StopIteration:
+            # input iterable was empty
+            yield from [] 
+        batch = []
+        for row in input:
+            if row["qid"] != thisqid or self.batch_size is not None and len(batch) == self.batch_size:
```
seanmacavaney, line 98: I think we can use `itertools.groupby` here to simplify a bit

cmacdonald:
I think this one is ready for re-review Sean.  Some notes for your consideration:

1. I've indeed reverted back to `.transform_iter()` returning `Iterable[Dict]`. This means it /can/ return a list (because `yield from List` works as expected), but returning a generator is supported, and indeed I think should encouraged. 
3. I wondered if the return type of `__call__` should be `List[Dict]`?
5. We may want to consider the default case, specifically whether `Transformer.transform_iter()` default impl needs to make a list from a DataFrame then return `iter()`. This is perhaps unnecessary?
7. pt.apply factories for functions those that take *multiple rows* can now accept an `iter=True` kwarg; no change is needed for single-row pt.apply factories. Test cases check that functions can return generators or lists.
9. pt.model.add_ranks support for iter-dicts is probably needed in the future.
10. Some typevars would be nice, but that's feature-creep at this stage.

pyterrier/apply_base.py
```diff
@@ -160,6 +194,16 @@ def __init__(self, fn,  *args, **kwargs):
 
     def __repr__(self):
         return "pt.apply.doc_features()"
+    
+    def transform_iter(self, iterdict):
+        fn = self.fn
+        # we assume that the function can take a dictionary as well as a pandas.Series. As long as [""] notation is used
+        # to access fields, both should work
+        def gen():
```
seanmacavaney, line 202: Why does this wrap an inner function instead of making gen the function itself? (same question for the ones below too)

pyterrier/ops.py
```diff
@@ -309,28 +309,34 @@ class ComposedPipeline(NAryTransformerBase):
...
         def gen():
             for batch in chunked(iter, batch_size):
-                batch_df = prev_transformer.transform_iter(batch)
-                for row in batch_df.itertuples(index=False):
-                    yield row._asdict()
+                yield from prev_transformer.transform_iter(batch)
         return last_transformer.index(gen()) 
 
-    def transform(self, topics):
+    def transform_iter(self, topics):
```
seanmacavaney, line 333: I don't understand the need for `_composed` here. Why not:
```
def transform_iter(self, inp):
   out = inp
   for transformer in self.models:
      out = transformer.transform_iter(out)
   return out
```

pyterrier/transformer.py
```diff
@@ -93,16 +101,28 @@ def transform(self, topics_or_res : pd.DataFrame) -> pd.DataFrame:
...
-            that avoids constructing a dataframe by hand. Alo used in the implementation of index() on a composed 
-            pipeline.
+            Method that proesses an iter-dict by instantiating it as a dataframe and calling ``transform()``.
+            Returns an Iterable[dict] equivalent to the DataFrame returned by ``transform()``, usually a list or a generator. 
+            This can be a handier version of ``transform()`` that avoids constructing a dataframe. Also used in the 
+            instantiation of ``index()`` on a composed pipeline.
         """
-        return self.transform(pd.DataFrame(list(input)))
+        # We should have no recursive transform <-> transform_iter problem, due to the __new__ check, UNLESS .transform() is called on an Indexer.
+        return iter(self.transform(pd.DataFrame(list(input))).to_dict(orient='records'))
```
seanmacavaney, line 115: Minor but I don't think we need to wrap in `iter` here, it's already iterable

pyterrier/transformer.py
```diff
@@ -93,16 +101,28 @@ def transform(self, topics_or_res : pd.DataFrame) -> pd.DataFrame:
...
+    
+    def __call__(self, input : Union[pd.DataFrame, Iterable[dict]]) -> Union[pd.DataFrame, Iterable[dict]]:
+        """
+            Sets up a default method for every transformer, which is aliased to ``transform()`` (for DataFrames)
+            or ``list(transform_iter())`` (for iterable dictionaries) depending on the type of input. The return type
+            matches the input type, but is always instantiated (generators should not be returned).
+        """
+        if isinstance(input, pd.DataFrame):
+            return self.transform(input)
+        return list(self.transform_iter(input))
```
seanmacavaney, line 125: I'd rather have `__call__` return whatever type you pass in. So if it's a dataframe, return as a dataframe, if it's a list, return it as a list, if it's a generic iterable, return a generic iterable.

pyterrier/apply.py
```diff
@@ -175,24 +179,49 @@ def generic(fn : Callable[[pd.DataFrame], pd.DataFrame], *args, batch_size=None,
...
+                for result in iterdict:
+                    if result["rank"] < 2:
+                        yield result
+            pipe1 = pt.terrier.Retriever(index) >> pt.apply.generic(_fn, iter=True)
+
+            # transform_iter() can also return an iterable, so returning a list is also permissible
+            pipe2 = pt.terrier.Retriever(index) >> pt.apply.generic(lambda res: [row for row in res if row["rank"] < 2], iter=True)
+
     """
+    if iter:
```
seanmacavaney, line 203: Needing an `iter` argument here is a bit unfortunate since it doesn't work the same way as the other applies.

pyterrier/apply_base.py
cmacdonald: groups returned by groupby are iterators, so we couldnt have batch_size support... I've used more_itertools.ichunked to resolve this.

pyterrier/apply.py
cmacdonald: its common across all that take more than one row as input...
a Series is close to a Dict
a DataFrame is not close to an IterDict

seanmacavaney: Yeah, I understand the necessity. The difference is something that might confuse users though.

pyterrier/transformer.py
cmacdonald: Not sure I'm as keen. `.__call__()` for me is user-facing magic for ease-of-use; anything trying to be clever that knows the type should call the correct `.transform()` or `.transform_iter()`? What would the sketch code look like?

cmacdonald:
Thx Sean, most comments addressed.

cmacdonald:
Question: can we somehow know how many times DataFrames are constructed, and check that this makes an indexing pipeline faster?

pyterrier/apply.py
cmacdonald: i think the extra functionality is hidden to most users, e.g. through iter=True additional kwarg. The overview table in apply.rst is a bit messy now though.

cmacdonald:
> We only told people to implement .transform() in the past, not .transform_iter()

I think the API change may be backwards compatible, at least in some cases. If .transform_iter() returns a DataFrame rather than an Iterable, it might work fine, at least in cases where the output of transform_iter() is used to construct a DataFrame (e.g. `pd.DataFrame(list(self.transform_iter()))` c.f. https://github.com/terrier-org/pyterrier/blob/transform_iter/pyterrier/transformer.py#L105))

If we really care about this, we could add some unit tests.

pyterrier/transformer.py
seanmacavaney: To me, there's an appeal to having a function that gives out whatever type is put in. If a user is savvy enough to use iterators, they hopefully know how they work.

> What would the sketch code look like?

```python
        if isinstance(input, pd.DataFrame):
            return self.transform(input)
        if isinstance(input, list):
            return list(self.transform_iter(input))
        # Are there other input types we want to handle specifically?
        return self.transform_iter(input)
```

seanmacavaney:
> I think the API change may be backwards compatible

I think you're right, and in the cases where it doesn't work, I'm happy with having it as a breaking change.

seanmacavaney:
> Question: can we somehow know how many times DataFrames are constructed, and check that this makes an indexing pipeline faster?

Perhaps we could patch `pandas.DataFrame` with one that counts the number of invocations? It depends on having transformers that implement `transform_iter()` though to have a positive effect though, right?

pyterrier/transformer.py
cmacdonald: Ok. Do you want to make the change (and to the types)?

cmacdonald:
> > Question: can we somehow know how many times DataFrames are constructed, and check that this makes an indexing pipeline faster?
> 
> Perhaps we could patch `pandas.DataFrame` with one that counts the number of invocations? It depends on having transformers that implement `transform_iter()` though to have a positive effect though, right?

At least an indexing pipeline should now be much simpler:
```python
pipeline = pt.apply.text(lambda row: row["title"] + " " + row["text"] )
pipeline.index(some_corpus_with_title_and_body)
```

pyterrier/transformer.py
cmacdonald: i.e return type of __call__ should now be:
```Union[pd.DataFrame, List[dict], Iterator[dict]]```
?

Also:
 - `dict` vs `Dict` vs `Dict[str,Any]`

**Comment by seanmacavaney** [comment]
Alright, @cmacdonald can you take a look at my change?

With another look, I'm not so jazzed that somebody who accidentally uses an `Indexer` as a regular transformer will get a stack overflow error. I'll try to address in another commit.

**Comment by seanmacavaney** [comment]
Alright, I think I sorted that out now too, with test cases to verify.

**Comment by cmacdonald** [comment]
Thanks Sean, this is looking really good now. I think we should:
(a) do some testing that this indeed improves indexing speed for some indexing pipelines, and 
(b) decide the next version number this applies to.

**Comment by seanmacavaney** [comment]
I tried and it was the same. Looks like `pt.apply.[x]` [still uses DataFrames](https://github.com/terrier-org/pyterrier/blob/transform_iter/pyterrier/apply.py#L251).

I mocked up a quick implementation below to see the potential gains, and it looks like we could see a ~2.5x speedup (98s for msmarco-document down to 37s).

```python
import pyterrier as pt

class NullIndexer(pt.Indexer):
  def index(self, it):
    for _ in it:
      pass

class NewGenericApply(pt.Transformer):
  def __init__(self, col, fn):
    self.col = col
    self.fn = fn
  def transform_iter(self, inp):
    for rec in inp:
      yield dict(rec, **{self.col: self.fn(rec)})

dataset = pt.get_dataset('irds:msmarco-document')

# iterator
pipeline = NewGenericApply('text', lambda x: '{title}\n{body}'.format(**x)) >> NullIndexer()
pipeline.index(dataset.get_corpus_iter()) # 37s (86345.29docs/s)

# dataframe
pipeline = pt.apply.text(lambda x: '{title}\n{body}'.format(**x)) >> NullIndexer()
pipeline.index(dataset.get_corpus_iter()) # 98s (32509docs/s)
```

cmacdonald:
Merged, thanks for your work on this too Sean :-)

seanmacavaney:
No problem, awesome enhancement!

PR #487, opened 2024-09-24
Artifact merged

cmacdonald:
updated artifact branch to current master


seanmacavaney:
A bunch of changes were made under alpha, so it might be easier to drop the existing branch and just pull from there instead.

Issue #489, opened 2024-09-24
 pyproject.toml 


cmacdonald:
DEPRECATION: Legacy editable install of python-terrier==0.11.0 from file:///Users/craigm/git/pyterrier (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

PR #490, opened 2024-09-25
move to pyproject.toml

cmacdonald:
closes: #489 

 - [x] how to override version for test releases to pypi
 - [ ] update wiki on how to do a release


cmacdonald:
> Perhaps sed `__init__.py` to modify the version in the source?

Thanks for the hint. This was somehow easier than trying to rewrite a the toml file, or find a way to override it.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+  "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)",
+  "Operating System :: OS Independent",
+]
+dynamic = ["version", "dependencies"]
+
+[tool.setuptools.dynamic]
+version = {attr = "pyterrier.__version__"}
+dependencies = {file = ["requirements.txt"]}
+
+#package_data={'': ['LICENSE.txt', 'requirements.txt', 'requirements-test.txt']}
```
cmacdonald, line 29: are these included?

cmacdonald: do we need requirements.txt included?

seanmacavaney: they don't need to be included

seanmacavaney (approved):

pyproject.toml
cmacdonald: you're the expert on this new format - can you make the changes?

PR #497, opened 2024-11-28
integrated extension documentation

seanmacavaney:


seanmacavaney:
The generated documentation can be found as an output of the `docs` github action here: https://github.com/terrier-org/pyterrier/actions/runs/12067562302/artifacts/2249221604

cmacdonald:
Like the new style. Some inelegancies are:
 - the index page just repeats the ToC
 - Query Formats for Terrier retrievers (terrier-retrieval.rst) doesnt render so well

>The generated documentation can be found as an output of the docs github action

Whats the purposes of this? Just for debugging doc generation? NB: it has QUICK=1, so some of the tables are outdated.

seanmacavaney:
> the index page just repeats the ToC

This is no different than the current documentation, no? I think the landing page should be improved, but that's a bit outside the scope of this PR.

image

> Query Formats for Terrier retrievers (terrier-retrieval.rst) doesnt render so well

Good catch, fixed. Also sorted several other rendering issues throughout. (Mostly due to lists being treated as block quotes.)

> Whats the purposes of this? Just for debugging doc generation? NB: it has QUICK=1, so some of the tables are outdated.

Yep, a build check to make sure that the documentation compiles. It outputs the generated documentation so it's easier for others to check (without needing to fetch and run themselves).

I think QUICK=1 is reasonable here, since we don't want the build step to take too long to run.

cmacdonald:
> I think the landing page should be improved, but that's a bit outside the scope of this PR.
fair

PR #498, opened 2024-12-02
Auto-generated Citations from DBLP

seanmacavaney:
Previously, references were pretty ad-hoc in the documentation, with inconsistent formatting, etc. E.g.,

image
image
image

This PR adds a new `.. cite.dblp::` directive that auto-generates citations from a provided DBLP key. for instance,

```
.. cite.dblp:: conf/sigir/MetzlerC05
```

image

This makes it easier to include references (since all you need is the DBLP key), makes the references consistently formatted, and gives a few features (like expanding for the bibtex and providing a link to the paper). A future version could also compile them all together for a bibliography page.

The DBLP bibtex that powers this is cached to avoid adding excessive time to the documentation build process.


cmacdonald (approved):

cmacdonald:
Amazing!

PR #501, opened 2024-12-04
craigs suggestions for compile branch

cmacdonald:
[cmacdonald](https://github.com/cmacdonald) commented [2 days ago](https://github.com/terrier-org/pyterrier/pull/480#issuecomment-2511409942):
 * _I wonder if ops.py should be renamed _ops.py? The operator transformer implementation arent really expected to be public_
 * _The SupportsFuseLeft etc protocols should be public though - i would think they should be in transformer.py and exposed in __init__.py?_


seanmacavaney (approved):

seanmacavaney:
thanks!

PR #506, opened 2024-12-10
Compilation improvements

cmacdonald:
- fix an assertion in feature-union
 - fix fuse_right compilation and add additional tests
 - fix FeaturesRetriever compilation that equates IndexRef objects

### DIFF
--- a/pyterrier/_ops.py
+++ b/pyterrier/_ops.py
@@ -381,7 +381,7 @@ class Compose(NAryTransformerBase):
 
     def compile(self, verbose: bool = False) -> Transformer:
         """Returns a new transformer that iteratively fuses adjacent transformers to form a more efficient pipeline."""
-        # compile constituent transformers (flatten allows complie() to return Compose pipelines)
+        # compile constituent transformers (flatten allows compile() to return Compose pipelines)
         inp = deque(_flatten((t.compile() for t in self._transformers), Compose))
         out : deque = deque()
         counter = 1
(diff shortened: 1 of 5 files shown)


seanmacavaney:
I'm puzzled on why the windows (py3.12) build would fail for this, but other settings pass.

cmacdonald:
It seems that compilation did not occur:
```
AssertionError: False is not true : compiled is not a FeaturesRetriever, it was (TerrierRetr(BM25) >> (TerrierRetr(DPH) ** TerrierRetr(PL2)) >> RankCutoff(2)) for indexloc <org.terrier.structures.Index at 0x221304dc810 jclass=org/terrier/structures/Index jself=<LocalRef obj=0x2f53e060 at 0x22162710dd0>>
```

The Retrievers were instantiated with an Index object. There would be a DirectIndexRef for each, which would should be identical (and would be equated).

I wonder if this is something to do with [DirectIndexRef.equals()](https://github.com/terrier-org/terrier-core/blob/5.x/modules/index-api/src/main/java/org/terrier/structures/Index.java#L104-L118).

cmacdonald:
```
indexref D:\a\pyterrier\pyterrier\tests\fixtures\index/data.properties not the same as indexref D:\a\pyterrier\pyterrier\tests/fixtures/index/data.properties
```
back-slashes vs forward-slashes -- ugh?

cmacdonald:
The failure was caused by a mismatch of opening indices using Index vs index location, on Windows, mixing / and \. Its so obscure, not going to fix.

PR #508, opened 2024-12-12
Py tokeniser - WIP

cmacdonald:


seanmacavaney:
The EnglishTokenizer implementation was around 4x slower than the Java one, so I took a stab at improving it using regular expressions (https://github.com/terrier-org/pyterrier/pull/508/commits/aaf486e9751ea57731b2cd6713dab7993ad6c6a1). This brought up throughput to about the same as the Java one.

I think we should have a test case that directly compares the tokenization of the java one vs the python one, over a large amount of text. It can be disabled by default, but it's important that we get these equivalent. There are probably subtleties to how each language handles things like UTF characters, etc.

cmacdonald:
Reasonable suggestion. Will work up something.

seanmacavaney:
Alright, both the English and UTF tokenizers are only around 20% slower than the Java versions and they match the tokenization on the first million documents from msmarco-passage.

I'm not certain what else you had in mind for this PR, but I've now addressed the bits that I was concerned about.

cmacdonald:
My motivation was only to untie dataset.get_topics() from Java. This sounds sufficient?

cmacdonald:
(which makes me wonder if pyterrier/terrier/tokeniser.py is the correct place for this)

seanmacavaney:
This seems like a reasonable place for them, given that they replicate Terrier's tokenizer behaviors, which is not necessarily what we'd suggest in general. (For example, the UTFTokeniser skips everything above U+FFFF, which will drop lots of (but not all) emoji, etc.)

Perhaps it's worth renaming them to avoid confusion with the Java versions? E.g., PyEnglishTokeniser?

I think we also need to add docstrings, update documentation, etc. to explain their motivation, how they should be used, etc.

cmacdonald:
> (which makes me wonder if pyterrier/terrier/tokeniser.py is the correct place for this)

If pyterrier.terrier is removed from this repo (into a separate repo), we'd need to move all this code.

cmacdonald:
@seanmacavaney Sean do you want to give another look?
I have made terrier.(Features)Retriever accept and parse queries it wouldn't previously have accepted. QE still works as expected. There is no change for an MSMARCO BM25 run: 
```python
from pyterrier.measures import *
pt.Experiment(
    [bm25], 
    pt.get_dataset('irds:msmarco-passage/trec-dl-2019/judged').get_topics(),
    pt.get_dataset('irds:msmarco-passage/trec-dl-2019/judged').get_qrels(),
    [AP@1000, NDCG@10]
)
```
```
                name   AP@1000  nDCG@10                                                                             
0  TerrierRetr(BM25)  0.370004  0.47954
```
What I think we need to do is see the impact on:
 - cross-encoder models
 - dense retrieval models

pyterrier/terrier/retriever.py
```diff
@@ -13,7 +14,30 @@
...
+    
+    if "applypipeline" in query:
+        return False
+    
+    # does it contains in terrier language 
+    termweightsre = re.compile(r'\^\d+(\.\d+)?')
+    if termweightsre.match(query):
+        return False
+
+    # we dont include : (or ^) in this list
```
seanmacavaney, line 30: Can the comment say why?

pyterrier/terrier/retriever.py
```diff
@@ -13,7 +14,30 @@
 from pyterrier.terrier.stopwords import TerrierStopwords
 from pyterrier.terrier.tokeniser import TerrierTokeniser
 
+def _query_needs_tokenised(query: str) -> bool:
```
seanmacavaney, line 17: This feels a bit brittle, but maybe okay?

The way I did it in pyterrier-anserini is to have multiple fields for the query. `query` representing the natural-langauge query, and `query_lucene` for the lucene query langauge version of the query. (This mimics how we handle the bow `query_toks` type of query, too).

This would require changes throughout several components, but I guess this is the type of breaking change we can do in a 1.0 release.

pyterrier/datasets/_core.py
```diff
@@ -519,8 +517,8 @@ def get_dataset(name: str, **configure_kwargs) -> Dataset:
         raise TypeError(f'Unsupported keyword arguments passed to get_dataset: {get_dataset}')
 
     # Temporary handling of topic tokenization
-    if not hasattr(result.get_topics, '_has_add_tokenize_query_arg_applied'):
-        result.get_topics = add_tokenize_query_arg(result.get_topics)
+    # if not hasattr(result.get_topics, '_has_add_tokenize_query_arg_applied'):
+    #    result.get_topics = add_tokenize_query_arg(result.get_topics)
```
seanmacavaney, line 521: Maybe we should keep the kwarg and give a warning? Otherwise there's a good chance that a lot of code will break.

pyterrier/terrier/retriever.py
cmacdonald: i know you are a proponent of this approach - I don't feel ready for this yet.

pyterrier/terrier/tokeniser.py
```diff
@@ -33,4 +35,96 @@ def _to_class(this) -> str:
...
+        raise ValueError(f'Unknown/unsupported tokeniser: {this}')
+
+class BaseTokeniser():
+    LOWERCASE = True
+    maxNumOfSameConseqLettersPerTerm = 3
+    maxNumOfDigitsPerTerm = 4
+    MAX_TERM_LENGTH = 20
+
+
+class UTFTokeniser(BaseTokeniser):
```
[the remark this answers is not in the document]
cmacdonald: probably you are right. move them to a gist?

cmacdonald:
NB: as get_topics() for SGML formatted TREC-topics still requires Java, this kinda has a bearing on #563.
I guess most people using PyTerrier for eg Robust are probably going to want terrier?

pyterrier/datasets/_core.py
```diff
@@ -517,10 +504,11 @@ def get_dataset(name: str, **configure_kwargs) -> Dataset:
         result._configure(**configure_kwargs)
     elif configure_kwargs:
         raise TypeError(f'Unsupported keyword arguments passed to get_dataset: {get_dataset}')
-
+    
+    # TODO: Sean what to put here?
```
cmacdonald: what to do here Sean?

requirements.txt
```diff
@@ -17,3 +17,4 @@ statsmodels # used for multiple testing correction
 # others
 dill # used for serialisation that is better than pickle, e.g. passing wmodel functions to terrier.Retriever
 lz4 # used for compression
+beautifulsoup4 # used for parsing SGML formatted TREC topics
```
seanmacavaney: Let's leave this as optional, given its very narrow use case?

cmacdonald: Fair comment; I think we leave as is and revisit this once we have other optional dependencies up and running.

PR #512, opened 2024-12-16
Dataset Decompression Improvements

cmacdonald:
comments as per commit messages

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -11,6 +11,7 @@ import requests
 from .io import autoopen, touch
 import pyterrier as pt
 import tarfile
+import zipfile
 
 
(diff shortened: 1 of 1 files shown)


seanmacavaney:
What's the motivation for delaying the decompression? From what I can tell, it's to allow the pgoress bar? At least for compressed tarfiles, this has the side-effect of needing to decompress the entire file twice, which can get pretty expensive.

I also suspect that tarfile is smart enough that it won't re-scan the entire file if the next record the requested file. But This might be worth double checking.

IIRC, zip files don't have either of these problems, due to how the file metadata is stored.

seanmacavaney:
Ahhh, okay, now I get what this is doing. Yes, this is way better than before.

seanmacavaney (approved):

PR #513, opened 2024-12-17
support verbose=True on by_query iter

cmacdonald:
We'd like to be able to have `verbose=True` on `@pta.transform.by_query()` decorator, i.e. `pt.apply.by_query(, iter=True, verbose=True)`

This patch is a possible version to allow that without materialising inp.

Here's the output of a test case:
```
>>> pt.apply.by_query(lambda iter: (i for i in iter), iter=True, verbose=True)([{'qid': 'q1'}, {'qid': 'q2'}])
pt.apply.by_query(): 100%|███████████████| 2/2 [00:00<00:00, 27060.03it/s]
[{'qid': 'q1'}, {'qid': 'q2'}]
```

### DIFF
--- a/pyterrier/apply_base.py
+++ b/pyterrier/apply_base.py
@@ -198,8 +198,7 @@ class ApplyIterForEachQuery(pt.Transformer):
         return "pt.apply.by_query()"
 
     def transform_iter(self, inp: pt.model.IterDict) -> pt.model.IterDict:
-        import collections.abc
-        if self.verbose and isinstance(inp, collections.abc.Sized):
+        if self.verbose:
             inp = pt.tqdm(inp, desc="pt.apply.by_query()")
         if self.batch_size is not None:
             for _, group in itertools.groupby(inp, key=lambda row: row['qid']):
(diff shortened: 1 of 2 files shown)


pyterrier/apply_base.py
```diff
@@ -190,12 +191,16 @@ def __init__(self,
...
         self.fn = fn
+        self.verbose = verbose
         self.batch_size = batch_size
 
     def __repr__(self):
         return "pt.apply.by_query()"
 
     def transform_iter(self, inp: pt.model.IterDict) -> pt.model.IterDict:
+        import collections.abc
+        if self.verbose and isinstance(inp, collections.abc.Sized):
```
seanmacavaney, line 202: I don't necessarily see a problem showing a progress bar even if it isn't sized -- tqdm will just show the throughput, which is still helpful.

But I'm happy to go either way on this.

seanmacavaney (approved):

cmacdonald:
I agree. What about putting tqdm on the input to groupby - does that make sense to you?

seanmacavaney:
yep, perfect

PR #517, opened 2024-12-18
Datasets via entry points

seanmacavaney:
This work-in-progress allows datasets to be defined in external packages through the entry point `pyterrier.dataset_provider`.

The existing datasets have been moved into two providers: `builtin` (for default set of datasets), and `irds` (for the ones provided by ir-datasets).

This sets the stage for a few things:
 - Allowing other packages to provide datasets without the current workaround (first import the package, then have the import add the datasets to `pyterrier.datasets.DATASET_MAP`).
 - Dropping the dependency to ir-datasets (by allowing it to provide the pyterrier-compatible interface itself)

A side benefit is that dataset objects don't need to be created until needed. This could reduce the import time of the core pyterrier package.

The diff looks messy due to moving stuff around. The key new bits are in `pyterrier/datasets/_core.py`.


seanmacavaney:
Thanks! Is this ready for merging?

cmacdonald:
Yes. pyterrier-rag should continue to function using DATASET_MAP right until I merge the other PR? (as I cant merge the other PR until there is a pyterrier release).

seanmacavaney:
yup, the change is backwards-compatible with DATASET_MAP.

cmacdonald:
merged, thanks Sean. I think we need a little Wiki page that we can point to for people wanting to add their own datasets. @asash has asked about this.

seanmacavaney:
Good idea, I've added details in #547

PR #519, opened 2024-12-19
bibliography file in documentation

seanmacavaney:
Automatically generate a bibliography file in the documentation. It links back to the pages where the citations were made.


cmacdonald:
This is cool, while also pointing out we need to add more citations to our own work. I wonder if we can (i) make the title come before the authors and (ii) extend the venue - it seems quite short...

seanmacavaney:
Thanks!

> (i) make the title come before the authors

This is easy if we also do the same for the citations blocks on the pages themselves (see below). It becomes a bit trickier (but still doable) if we want a different format in each setting.

If we change, we'll have to manually update the non-dblp citations we have to match, but that should be doable.

But what's the motivation for this change? ACM/ACL/LNCS all list authors before title.

image

> (ii) extend the venue - it seems quite short...

I'd like to stick with what we can easily get from DBLP to avoid complicating the feature too much. DBLP gives us either "condensed" (e.g., _"SIGIR"_) or "standard" (e.g., _"Proceedings of the 47th International ACM SIGIR Conference on Research and Development in Information Retrieval, SIGIR 2024, Washington DC, USA, July 14-18, 2024"_). Between the two, I'd much prefer condensed.

I think most of those using the documentation will be familiar with the abbreviated venue names. And there's always the DOI link, which should give more complete bibliometric information for those looking to disambiguate further.

cmacdonald:
ok so anything else preventing merging this?

seanmacavaney:
I think it's good to merge if you're happy with it!

Issue #482, opened 2024-09-18
static analysis tools


seanmacavaney:
Let's add static analysis tools, like flake8/mypy/ruff/etc into the automatic build.

The questions are:
 - Which tools do we want to use?
 - Which rules?

Maybe we also want to make sure that the docs also build without errors (in `QUICK=1` mode) as another check?

Context:
We currently test a few custom rules using a flake8 extension: `extras/pyterrier-flake8-ext` config is here: https://github.com/terrier-org/pyterrier/blob/master/.github/workflows/style.yml

It looks like we used flake8 previously, but it was commented out? https://github.com/terrier-org/pyterrier/blob/master/.github/workflows/push.yml#L73C13-L79

I've been using ruff with [this config](https://github.com/seanmacavaney/pyterrier-anserini/blob/main/ruff.toml) in a few extension packages

cmacdonald:
> It looks like we used flake8 previously, but it was commented out? https://github.com/terrier-org/pyterrier/blob/master/.github/workflows/push.yml#L73C13-L79

It was never used really. I think it was copied over from another project's yml file.

One thing with flake, mypy etc is how to detect /changes/ - i.e. more errors than the baseline.  Simplest thing I can think of is somehow storing the most recent mypy output in .github folder, and doing a diff?  We'd have to work out how to see if the error count was reduced though

seanmacavaney:
> One thing with flake, mypy etc is how to detect /changes/ -

I think I disagree here -- we find a set of rules that need to be adhered to throughout the codebase. Then there's no notion of a delta. (`# noqa` can be used in cases of exceptions to the rule.)

seanmacavaney:
This is partially addressed in #484 with the addition of mypy to the style action.

I propose we also add ruff (with default ruleset), and then we have good coverage.

seanmacavaney:
The `mypy` action currently doesn't fail when errors are identified. This makes it easy to miss new errors introduced. I think triggering a failure of the action on errors and addressing any outstanding mypy errors should be part of this issue.

cmacdonald:
i dont think all mypy errors are resolvable currently - anything involving Jnius is a nightmare. I tried writing stubs for it, to no avail.

seanmacavaney:
[...]

PR #523, opened 2025-01-04
ruff github action

seanmacavaney:
more progress towards #482

### DIFF
--- a/pyterrier/datasets.py
+++ b/pyterrier/datasets.py
@@ -13,6 +13,7 @@ from .io import autoopen, touch
 import pyterrier as pt
 import tarfile
 import zipfile
+import ir_datasets
 
 
 TERRIER_DATA_BASE="http://data.terrier.org/indices/"
@@ -335,7 +336,7 @@ class RemoteDataset(Dataset):
     def _describe_component(self, component):
         if component not in self.locations:
             return None
-        if type(self.locations[component]) == type([]):
+        if isinstance(self.locations[component], list):
             return True
         if isinstance(self.locations[component], dict):
             return list(self.locations[component].keys())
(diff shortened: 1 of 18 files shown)


pyterrier/terrier/index_factory.py
```diff
@@ -60,17 +60,15 @@ def _load_into_memory(index, structures=['lexicon', 'direct', 'inverted', 'meta'
             }
 
         pindex = pt.java.cast("org.terrier.structures.IndexOnDisk", index)
-        load_profile = pindex.getIndexLoadingProfileAsRetrieval()
```
seanmacavaney, line 63: @cmacdonald, from what I can tell, this line can be removed. But I'm also not certain about what it was trying to do originally, so I'd appreciate it if you could double check this one.

cmacdonald (changes requested):

pyterrier/datasets.py
```diff
@@ -759,14 +760,6 @@ def msmarco_document_generate(dataset):
 MSMARCO_PASSAGE_FILES = {
     "corpus" : 
         [("collection.tsv", "collection.tar.gz#collection.tsv")],
-    "index": {
```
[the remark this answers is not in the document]
seanmacavaney: It wasn't doing anything -- the key "index" is defined (and therefore overwritten) below.

pyterrier/terrier/index_factory.py
seanmacavaney: Sorry, I'm still not seeing it -- can you point me to the specific line?

cmacdonald: looked at the wrong context, this occurrence is superfluous and can be removed.

Issue #533, opened 2025-01-15
TerrierIndex (artifact) API support loading indices into memory


cmacdonald:
We need to think about to how extend the API - the TerrierIndex constructor probably isnt the natural place...

seanmacavaney:
A few options off the top of my head:
 - Option A: Constructor `TerrierIndex(path, in_memory=True)`
    - Downside: 1 Not great when doing stuff like from_hf()
    - Downside 2: It's not really part of the definition of the artifact, it's a property of how you're using it
 - Option B: Method that returns a new index: `index.to_memory()` 
    - Addresses Downside A1 but A2 still applies
 - Option C: The transformers specify which version to use, e.g., `index.bm25(in_memory=True)` ; facilitated by something like an `index.index_ref_inmem`?
    - This matches the behavior of DR, in which the transformers are responsible for figuring out whether to load into memory or not
    - Addresses both downsides A1 and A2
    - Downside 3: This means we'd need to change the already hefty Retrieve stuff.
    - Downside 4: We'd have to make corresponding changes to other components that use it too, e.g., text loader, PRF, etc

cmacdonald:
C was what I was thinking; but it would just replace the index with an in-memory index. If you can load the index into memory for one transformer, you can keep it in memory for others.

seanmacavaney:
Right, since the retriever takes an index reference instead. Sounds reasonable.

Option A could work with from_hf() if we let it take kwargs and pass them through. But that's a bit ugly, and doesn't address Downside A2.

cmacdonald:
> Option A could work with from_hf() if we let it take kwargs and pass them through. But that's a bit ugly, and doesn't address Downside A2.

I'm ok with this tbh, as Hugggingface do the same, and from_dataset() does the same.

seanmacavaney:
I'm open to it as long as we're careful not to swallow kwargs in the artifact's constructor, please!

cmacdonald:
I have a similar problem about controlling metadata obtained from a Terrier index. Do you agree that `.bm25()` etc pass **kwargs to the TerrierRetriever?

seanmacavaney:
Nah, I really see this as an opportunity to fix a ton of the rough edges in the current API. If we don't share this vision, perhaps this API should be dropped from core and instead added as an extension package. What do you think?

cmacdonald:
kwargs are like a safety-valve - an override to fix things for advanced users that a new API doesnt support.

[...]

PR #534, opened 2025-01-15
Addresses #533 - kwargs for artifacts

cmacdonald:
See #533 for discussion

### DIFF
--- a/pyterrier/_artifact.py
+++ b/pyterrier/_artifact.py
@@ -41,7 +41,7 @@ class Artifact:
         self.path: Union[Path, _NoPath] = path if isinstance(path, _NoPath) else Path(path)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


seanmacavaney (commented):
Looks good, just a few minor comments.

cmacdonald:
> I think I'd prefer in_memory -- so as not be be confused with e.g., a buffer to load it into.

Unfortunately, I'd like it to match the name of the arg of the existing API, which is `memory`. 

NB: memory can be `Union[bool, List[str]]`

seanmacavaney:
`in_memory` intuitively works works when `List[str]`, too :)

`index.memory` also doesn't make sense to me as an attribute of an index/artifact. It sounds like it's a memory buffer or something, rather than a setting.

cmacdonald:
Your arguments are fair if an existing API doing the same thing didnt already exist....

PR #535, opened 2025-01-28
artifacts - push/pull from private huggingface repos

seanmacavaney:
Sometimes you may want to push/pull artifacts from private huggingface repositories. This PR adds support for this.


seanmacavaney:
From what I can tell, the build failures seem to be a jitpack issue, e.g., this url times out: https://jitpack.io/com/github/terrier-org/terrier-core/terrier-assemblies/5.x-SNAPSHOT/terrier-assemblies-5.x-SNAPSHOT-jar-with-dependencies.jar

cmacdonald:
its still timing out. One suggestion is separating the testing of jitpack snapshot versus all the OS versions (which would run using Maven released version). Perhaps jitpack is a separate yml file?

PR #536, opened 2025-01-30
use latest terrier in builds instead of snapshot

seanmacavaney:
jitpack seems to be having issues


cmacdonald:
we can consider adding another yml file for snapshot later...

Issue #538, opened 2025-02-15
Using Spanish custom Stopwords


AgustinNormand:
**Describe the bug**
During indexing, I am wanting to add a list of custom stopwords.
If I run the indexing without the list, it works fine.
If I add the list, I get an error.

Additionally, I believe there is a typo in the documentation,
Shouldn't the stopwords parameter be of type "TerrierStopwords"?

<img width="746" alt="Image" src="https://github.com/user-attachments/assets/89eaaee1-c9ad-4b3e-9c00-5d119e459ded" />

**To Reproduce**
Im using python-terrier 0.13.0
Define a File Indexer as:
```
indexer = pt.FilesIndexer(index_path = OUTPUT_DIR, 
                          overwrite = True, 
                          verbose = True,
                          stemmer = config["SPARSE_STEMMER"],
                          tokeniser = config["SPARSE_TOKENIZER"],
                          stopwords = ['a', 'an', 'the'])

indexref = indexer.index(INPUT_DIR)
```

Where in config:
SPARSE_TOKENIZER = "utf"
SPARSE_STEMMER = "SpanishSnowballStemmer"

Result:
```
Traceback (most recent call last):
  File "/Users/agustin/Git/digesto-hybrid-search/main.py", line 170, in <module>
    menu()
  File "/Users/agustin/Git/digesto-hybrid-search/main.py", line 167, in menu
    process_option(menu_entry_index)
  File "/Users/agustin/Git/digesto-hybrid-search/main.py", line 84, in process_option
    terrier_index()
  File "/Users/agustin/Git/digesto-hybrid-search/indexers/sparse_indexer/terrier_indexer.py", line 46, in terrier_index
    index_directory(documents_dir(config["SECTION_VISTO_DIR"]), config["SPARSE_INDEX_VISTO_DIR"], stopwords)
  File "/Users/agustin/Git/digesto-hybrid-search/indexers/sparse_indexer/terrier_indexer.py", line 29, in index_directory
    indexref = indexer.index(INPUT_DIR)
  File "/Users/agustin/Library/Caches/pypoetry/virtualenvs/digesto-hybrid-search-j6XPZ_0F-py3.10/lib/python3.10/site-packages/pyterrier/utils.py", line 208, in _wrapper
    return fn(*args, **kwargs)
[...]

PR #545, opened 2025-04-06
fix index hooks in FilesIndexer and DFIndexer (which implement custom stopword support)

cmacdonald:
In #538, @AgustinNormand noted a problem with custom stopwords for the FilesIndexer. The same problem exists in the DFIndexer.

### DIFF
--- a/pyterrier/terrier/index.py
+++ b/pyterrier/terrier/index.py
@@ -522,12 +522,20 @@ class DFIndexer(TerrierIndexer):
         self.index_called = True
         collectionIterator = None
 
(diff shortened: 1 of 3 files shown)


seanmacavaney (approved):
This looks reasonable, but I don’t quite follow how the cleanup hooks address the custom stopwords. Maybe a comment to explain a bit?

cmacdonald:
i think custom stopwords are written to the index properties using the cleanup hook. Is it enough to change the PR name?

seanmacavaney:
nah, just clarification on my end. lgtm

PR #546, opened 2025-04-06
replace wget dependency with pt.io.download

### LINKED ISSUE #522: replace wget with pt.io.download
Once #436 is merged, we can drop the wget dependency with the new `pt.io.download` function (as suggested by @cmacdonald https://github.com/terrier-org/pyterrier/pull/436#discussion_r1775610098)

cmacdonald:
fixes #522 

I tend to think that some of the mavenresolver logic could be simpler now - its trying to give alternative filenames etc.

### DIFF
--- a/pyterrier/java/mavenresolver.py
+++ b/pyterrier/java/mavenresolver.py
@@ -4,7 +4,7 @@ from packaging.version import Version
 from pathlib import Path
 from enum import Enum
 from warnings import warn
-import wget
+import requests
 import os
 import re
 import pyterrier as pt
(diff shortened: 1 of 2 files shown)


seanmacavaney (approved):
awesome, looks great.

I think there's a slight difference in behavior. E.g., wget will avoid replacing a file that already exists by appending something like ` (#)` to the end of the file name if one already exists. But AFAIK we never rely on that behaviour, so I think we're good.

cmacdonald:
I think it could perhaps be simplified, but lets see how this does. I checked this works offline and online

PR #547, opened 2025-04-06
Documentation for Extending PyTerrier (starting point)

seanmacavaney:
Here's a starting point for documentation on extending PyTerrier. It includes stuff on adding datasets (as suggested in #517) and some basic details and on creating extension packages.

This is something we should continue building up going forward (e.g., writing an extension Transformer, Artifact, etc.).


cmacdonald:
Thanks for bashing this out.

Two things:

1. I think perhaps we need to give a simpler example - say I make a new TREC dataset and its not in IRDS, can we link to page that gives a very short description of how to bring it in?

2. Want to bring in this guide on LSR backends?
https://gist.github.com/cmacdonald/d0160a72490b99d0ed6e85b5b3a6c4f5

seanmacavaney:
> need to give a simpler example

I think in that case, we'd actually recommend that people add it to ir_datasets instead. This is for more exotic stuff that doesn't necessarily fit the typical queries/docs/qrels paradigm (like the stuff in pyterrier_rag).

I'll add notes on this.

> Want to bring in this guide on LSR backends?

Sure, thanks!

cmacdonald:
> I think in that case, we'd actually recommend that people add it to ir_datasets instead. This is for more exotic stuff that doesn't necessarily fit the typical queries/docs/qrels paradigm (like the stuff in pyterrier_rag).

All agreed - to be clear, three use cases:
(1) use my own data
(2) make my own data accessible via pt.get_dataset()
(3) make my own data publicly accessible via IRDS

seanmacavaney:
Yup! There's a bit on the note at the top. Is that clear enough?

seanmacavaney:
As an aside, stuff is in the works to make it a bit simpler to add to ir-datasets (with an accompanying guide).

cmacdonald (approved):

seanmacavaney:
> The link to the pt.Artefact documentation doesnt work

That's in this branch: https://github.com/terrier-org/pyterrier/tree/artifact-later, which we left out of the original merge to avoid confusing students. We're probably good to make a PR on this soon, now that the teaching period is over?

cmacdonald:
> That's in this branch: https://github.com/terrier-org/pyterrier/tree/artifact-later, which we left out of the original merge to avoid confusing students. We're probably good to make a PR on this soon, now that the teaching period is over?

yes, go for it.

If artifact-merged is merged, can you delete?

PR #551, opened 2025-04-23
add pyterrier_rag to extension docs

cmacdonald:


seanmacavaney:
Changing the line to
```
# RAG <https://github.com/terrierteam/pyterrier_rag>
```

will make the GHA pass, but only link to the repo.

Otherwise we can wait until it's on pypi.

PR #553, opened 2025-05-09
migrate push_queries

cmacdonald:
in https://github.com/terrierteam/pyterrier_rag/pull/24, a generic form of push_queries etc was introduced by @Parry-Parry. This merges those methods upstream.

At the same time, I've removed support for inplace editing of the dataframe.

### DIFF
--- a/pyterrier/apply_base.py
+++ b/pyterrier/apply_base.py
@@ -383,7 +383,7 @@ class ApplyQueryTransformer(pt.Transformer):
     def transform(self, inp: pd.DataFrame) -> pd.DataFrame:    
         if "query" in inp.columns:
             # we only push if a query already exists
-            outputRes = pt.model.push_queries(inp.copy(), inplace=True, keep_original=True)
+            outputRes = pt.model.push_queries(inp, keep_original=True)
         else:
             outputRes = inp.copy()
         try:
--- a/pyterrier/model.py
+++ b/pyterrier/model.py
@@ -2,7 +2,7 @@ import math
 import itertools
 import numpy as np
 import pandas as pd
-from typing import Any, Dict, Iterable, List, Sequence, Optional
+from typing import Any, Dict, Iterable, List, Sequence, Optional, Union, Tuple
 
 IterDictRecord = Dict[str, Any]
 IterDict = Iterable[IterDictRecord]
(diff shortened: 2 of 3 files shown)


cmacdonald:
mypy errors:
pyterrier/model.py:191: error: Incompatible return value type (got "map[Any]", expected "Tuple[str, int]")  [return-value]
pyterrier/apply_base.py:379: error: Incompatible types in assignment (expression has type "Union[Dict[str, Any], Iterable[Dict[str, Any]]]", variable has type "Dict[str, Any]")  [assignment]

pyterrier/model.py
```diff
@@ -85,80 +85,134 @@ def query_columns(df : pd.DataFrame, qid=True) -> Sequence[str]:
...
-        next_col = f'query_{query_idx}'
-        if prev_col in inp:
-            inp[next_col] = inp.pop(prev_col) # assign e.g., query_1 to query_0
-            prev_col = next_col
-        else:
-             break
-    if keep_original:
-        inp['query'] = inp['query_0']
-    return inp
+    def per_element(i: IterDictRecord) -> IterDictRecord:
```
seanmacavaney, line 147: We may want to try to make this a bit more efficient in the `IterDict` case, since it's doing a lot of work for every record. But it's not a regression, so not necessary to address in this PR.

pyterrier/model.py
```diff
@@ -85,80 +85,134 @@ def query_columns(df : pd.DataFrame, qid=True) -> Sequence[str]:
...
+
+def push_queries_dict(inp: Union[IterDictRecord,IterDict], keep_original: bool = False, inplace: bool = False) -> Union[IterDictRecord,IterDict]:
+    """
+    Works like ``push_queries`` but over a dict/iterdict instead of a dataframe.
+    """
+    assert not inplace, "push_queries_dict does not support inplace"
+    return push_columns_dict(inp, keep_original=keep_original, base_column="query")
+
+
+def pop_columns(df: pd.DataFrame, base_column="query") -> pd.DataFrame:
```
seanmacavaney, line 206: Are we missing a pop for the dict case? This also isn't a regression, just surprising.

pyterrier/apply_base.py
```diff
@@ -376,14 +376,14 @@ def transform_iter(self, inp: pt.model.IterDict) -> pt.model.IterDict:
         for row in inp:
             row = row.copy()
             if "query" in row:
-                row = pt.model.push_queries_dict(row, inplace=True, keep_original=True)
+                row = pt.model.push_queries_dict(row, keep_original=True)
```
seanmacavaney, line 379: This should be inplace to avoid making an extra copy, right?

seanmacavaney (approved):
Could probably use a few unit tests for `push_columns` and `pop_columns` directly.

pyterrier/model.py
cmacdonald: yes

pyterrier/apply_base.py
cmacdonald: i disabled a lot of inplace=True, to avoid the temptation for downstream users. 

however, there are places where it used to make sense, e.g. if a row was new anyway. thoughts?

seanmacavaney: I don't feel particularly strongly

PR #554, opened 2025-05-09
Show types of arguments in docs

cmacdonald:
It turns out that Arguments with a list will not propagate types. :param notation does. This updates a lot of the documentation to use :param etc.

WIP

### DIFF
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Callable, Any, Dict, Union, Optional, Sequence
+from typing import Callable, Any, Dict, Union, Optional, Sequence, Literal
 import numpy.typing as npt
 import pandas as pd
 import pyterrier as pt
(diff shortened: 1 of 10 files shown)


seanmacavaney:
I think setting `autodoc_typehints = 'description'` or `autodoc_typehints = 'both'` (depending on whether we want typehints to also appear in the signature) can accomplish this without needing to change the docstring format.

cmacdonald:
> I think setting `autodoc_typehints = 'description'` or `autodoc_typehints = 'both'` (depending on whether we want typehints to also appear in the signature) can accomplish this without needing to change the docstring format.

What I tried was that they only appeared in the signature, which seemed ugly, and didnt get rid of the problem of incomplete or wrong manual hints in the argument list. I think most of the work is now done, just trying to address some mypy teething errors

seanmacavaney:
Alright. Can we still set `autodoc_typehints = 'description'` to support type hints in extensions? I'd rather not have to go through them all. (Selfishly, I also prefer how the "Arguments:"-style over the ":param"-style visually.)

This is how it appeared when I tried it, which seems fine.

image

cmacdonald:
there are also other minor documentation improvements, added some types for pt.terrier.Retriever etc.

cmacdonald:
> Alright. Can we still set `autodoc_typehints = 'description'` to support type hints in extensions? I'd rather not have to go through them all. (Selfishly, I also prefer how the "Arguments:"-style over the ":param"-style visually.)

I had this on when developing this PR.

PR #557, opened 2025-05-14
fix(typo): docstring of IterDictIndexer

iqbalpa:
Fixed typo from `tranfer` to `transfer` in IterDictIndexer's docstring


seanmacavaney:
Thanks @iqbalpa!

PR #563, opened 2025-08-15
pyterrier where java is optional...

cmacdonald:
PoC - requires discussion as to what should be the default setting


seanmacavaney:
Between the two, I think I'd lean towards having a separate `core` package, since this would avoid breaking existing dependencies on `python-terrier` that may assume it includes terrier functionality.

Issue #570, opened 2025-08-19
inspect support for pt.apply.doc_features, pt.apply.doc_score, pt.apply.generic, pt.apply.by_query


cmacdonald:
Test cases needed for inspection of much of pt.apply...

PR #572, opened 2025-08-21
addresses #570 - testing and refining inspectability of pt.apply transformers; inspectability of transform_iter

cmacdonald:
addresses: #570

This PR:
 - ensures that the various pt.apply functions are inspectable
 - makes Transformers that only implement transform_iter() uninspectable defacto unless transform_outputs is defined.

### DIFF
--- a/pyterrier/apply.py
+++ b/pyterrier/apply.py
@@ -190,8 +190,9 @@ def generic(
         :param verbose: Whether to display a progress bar over batches (only used if batch_size is set, and iter is not set).
         :param iter: Whether to use the iter-dict API - if-so, then ``fn`` receives an iterable, and returns an iterable. 
         :param transform_outputs: Used to support inspection. If provided, should be a function (e.g. lambda) that takes 
-            self and input columns as input arguments, and return a list of columns that the transformer will output. 
-            This need only be set if you have iter=True, or your transformer doesn't respond well to empty dataframes.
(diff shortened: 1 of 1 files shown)


pyterrier/apply.py
```diff
@@ -205,10 +214,21 @@ def _fn(iterdict):
...
             raise ValueError("add_ranks=True not supported with iter=True")
-        return ApplyGenericIterTransformer(fn, *args, batch_size=batch_size, **kwargs)
-    return ApplyGenericTransformer(fn, *args, batch_size=batch_size, **kwargs)
-
-def by_query(fn : Union[Callable[[pd.DataFrame], pd.DataFrame], Callable[[pt.model.IterDict], pt.model.IterDict]], *args, batch_size=None, iter=False, verbose=False, **kwargs) -> pt.Transformer:
+        rtr = ApplyGenericIterTransformer(fn, *args, batch_size=batch_size, **kwargs)
+    else:
+        rtr = ApplyGenericTransformer(fn, *args, batch_size=batch_size, **kwargs)
+    if transform_outputs is not None:
+        _bind(rtr, transform_outputs, as_name='transform_outputs')
```
seanmacavaney, line 224: Are you sure we need bind here? Can't we pass transform_outputs straight through as a kwarg?

I guess that means it wouldn't take self as an argument, but I don't think you ever really need a reference back to self for an apply.

PR #573, opened 2025-08-21
upgrade minimum version of Python to 3.9

cmacdonald:
Py 3.8 is EOL
We should also be testing for Py 3.13 now


cmacdonald:
3.13 not possible yet until jnius does a release

PR #575, opened 2025-08-22
refactor frame_info out of schematic.py

cmacdonald:


seanmacavaney:
Looks good, can it return a dict instead like `column_info` so that we can maybe add stuff to it in the future?

cmacdonald:
revised

PR #577, opened 2025-08-28
Funion inspect

cmacdonald:
This PR ensures that transform_inputs identifies the minimum subset of columns that can work with a given feature union


cmacdonald:
test failure is a Python 3.8 thing, which we no longer support. master has a PR to move to 3.9 minimum

cmacdonald:
sure; I thought that other ops might need it, hence why i put it in that class. any suggestions for a name?

PR #579, opened 2025-09-03
Schematic docs can specify input cols

cmacdonald:
This fixes an issue I noted in the terrier-retrieval.rst where the reranker was not being shown as such.

image


seanmacavaney (approved):
Looks great, thanks!

PR #580, opened 2025-09-04
schematic doc fixes for pt1

cmacdonald:
This PR makes various fixes to add more schematics to the core docs.

 - [x] ltr.rst - sub DF dont fit in the box
 - [x] ltr.rst - why dont we know the output types
 - [x] ops.rst - can we add more examples here.


seanmacavaney:
> 2 problems in ltr.rst

Thanks, sorted!

cmacdonald:
> fix rendering of inner dfs with 2-character labels
yes, this fixes it visually, but I was thinking we should be able to reason on the types of doc_features and doc_score.

PR #582, opened 2025-09-08
start testing py3.13

cmacdonald:
(because Jnius 1.7.0 has been released, which we also need to test with)


seanmacavaney:
> Jnius 1.7.0 has been released

🎉

PR #586, opened 2025-09-24
allow inspection of transformer without inspectable attributes

cmacdonald:
Adds a strict kwarg for pt.inspect.transformer_attributes, and uses that in subtransformers.


seanmacavaney (approved):
lgtm, I changed from `...` to a unique value. `transformer_apply_attributes` then checks this and fails if it's present.

PR #599, opened 2025-10-16
Fit: adding title property in the dictionary that is being returned b…

lionisakis:
…y the column info


cmacdonald:
thanks. Remind us why this is useful?

PR #604, opened 2025-10-31
Fix Colab without java

cmacdonald:
As noted in https://github.com/googlecolab/colabtools/issues/5686, Google Colab has unexpectedly removed Java. 
This PR has a workaround in the form of a new initializer. We can merge if this becomes a long-standing issue. We may even backport to 0.13.


cmacdonald:
We've release 0.13.2 with this PR included.

PR #613, opened 2025-11-14
Tweaks to artifact and indexer schematics

seanmacavaney:
- Label indexer outputs as "Artifact" instead of "Output"
 - Don't render schematics for lone artifacts in notebooks


cmacdonald:
ta

PR #614, opened 2025-11-14
autotokenise for SDM, Bo1 etc.

cmacdonald:
Since 1.0, while terrier.Retriever applies autotokenisation to badly formatted queries, we didnt do the same for the rewriting classes in .terrier.

### DIFF
--- a/pyterrier/terrier/rewrite.py
+++ b/pyterrier/terrier/rewrite.py
@@ -9,8 +9,8 @@ from pyterrier.terrier.index import TerrierTokeniser
 def tokenise(tokeniser : Union[str,TerrierTokeniser,FunctionType] = 'english', matchop=False) -> pt.Transformer:
     """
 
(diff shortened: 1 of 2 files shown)


pyterrier/terrier/rewrite.py
```diff
@@ -179,7 +186,16 @@ class QueryExpansion(pt.Transformer):
...
-    def __init__(self, index_like, fb_terms=10, fb_docs=3, qeclass="org.terrier.querying.QueryExpansion", verbose=0, properties={}, requires_scores=False, **kwargs):
+    def __init__(self, 
+                 index_like, 
+                 fb_terms=10, 
+                 fb_docs=3, 
+                 qeclass="org.terrier.querying.QueryExpansion", 
+                 verbose=0, 
+                 properties={}, 
+                 requires_scores=False, 
+                 tokeniser : Union[str,TerrierTokeniser] = TerrierTokeniser.english,
```
seanmacavaney: I think it should be able to get the tokeniser from `index_like`?

PR #620, opened 2025-12-03
make space for all dependencies in docs GHA

cmacdonald:
docs.yml GHA was running out of space; this PR fixes the GHA


seanmacavaney:
Nice, that this saves loads of space. I suspect most of the dependency space is used for stuff like torch, so this should last us a good while.

PR #624, opened 2026-01-22
Fix warning message for offline maven access

seanmacavaney:
Not all URL Errors have `code` (e.g., "Network Unreachable"). https://docs.python.org/3/library/urllib.error.html

This causes an error to be raised in generating the warning.

@Parry-Parry


cmacdonald:
image

has these been an irds release?

PR #637, opened 2026-03-29
Py3.14

cmacdonald:


cmacdonald:
Need zlib_state release for Windows under Python 3.14.

PR #638, opened 2026-03-29
Add context in the form of the calling Transformer to pt.validate error message

### LINKED ISSUE #632: InputValidationError should say which transformer

cmacdonald:
This PR adds a context kwarg to all of the validation - this allows the calling code to provide context to the error message, which can be useful to non-experienced readers seeing an InputValidationError

closes: #632


seanmacavaney (approved):
It feels a little odd to pass something that isn't directly used (only for inclusion in the error message). But seems helpful and worth including. Lgtm.

PR #646, opened 2026-04-14
missing kwargs passthrough in Artifact loading

cmacdonald:


cmacdonald:
Sean and I discussed offline; we'll back out this change.

cmacdonald:
This is needed for pyterrier_colbert, as we agreed last week its really difficult to rearchitect its Artifact not to have constructor parameters

PR #648, opened 2026-05-09
Trie Execution Plan and Visualisation

irene-anu:
Trie-based execution plan for experiment. This is a new draft for feedback reflecting the right files and folders.

Couple of things I noticed that need to be fixed:
- [x] The visualisation renders differently on VSCode vs Google Colab (red/green/yellow not translated as expected in Colab)
- [ ] disable type checking colours for pipelines that dont inspect correctly.
- [x] Hover Infobox are not rendering, and causing JS errors.
-  [x] A few more use cases need to be integrated into current visualisation.
- [x] Change experiment test cases such that they all run with plan=tree too.
- [x] Check and test the commandline mode pretty print
- [x] Documentation in experiments.rst should be updated.

PR #652, opened 2026-05-20
fix: mrt now returns mean retrieval time per query

### LINKED ISSUE #627: mrt is not an average
**Describe the bug**
mrt is not an average

**Workaround**
Divide mrt by total number of queries

**Solution**
execution is returning a total time, which the render expects to be an average...

prisha2410:
Previously _run_and_evaluate() returned total runtime across all queries, causing mrt in pt.Experiment to display total time instead of mean time per query.

Fix: divide runtime by num_q before returning.

Fixes #627


cmacdonald:
Thanks for this @prisha2410. I'm going to pause this until #648 lands, as there is work ongoing there.

cmacdonald:
Thanks @prisha2410. I reduced some of the test cases, which seemed superfluous
@irene-anu and I also reviewed,  and integrated #648 related changes too

PR #674, opened 2026-07-25
Fix a dozen bugs, added tests

jegentile:
Correctness fixes:
- pipelines: Evaluate/GridSearch deprecation aliases wrapped the wrong functions (Experiment/GridScan); point them at Evaluate/GridSearch.
- terrier/retriever FeaturesRetriever: an empty query_toks did `return []`, aborting the whole batch and returning a list where a DataFrame is expected; use `continue`.
- terrier/retriever TextIndexProcessor: returns="queries" backfill called DataFrame.merge with no right frame; merge topics with inner_res on qid and coalesce missing rewritten queries (incl. query_N columns) from the original topic. Fix "docs of queries" message typo.
- text SlidingWindowPassager: short-doc branch called dict.drop(); use del to match the long-doc branch.
- Shared-frame mutation: SetUnion/SetIntersection/Sum/Concatenate now give each operand its own copy; tree-plan sibling branches and the default linear plan / GridScan copy topics per system, so a mutating transformer can no longer corrupt another system's input or the caller's frame.
- io: URL topic loading was broken. _NosyReader.read() with no size capped at one buffer instead of reading to EOF; wrap URL download streams in a TextIOWrapper so parsing sees text. letor append used invalid mode 'wa' (use 'at'). Singleline parser raised UnboundLocalError with contains_qid=False and silently duplicated the previous query on a malformed line; assign query in the no-qid branch and raise on non-match.
- apply.doc_score forwards required_columns, enabling its documented validation (skipped for zero-column empty-iterable input).
- java/_utils JavaClasses.__getattr__ returned an AttributeError instead of raising it, breaking hasattr and yielding confusing downstream errors.
- __init__: `from pyterrier import *` always failed; stub ApplicationSetup and drop the never-defined 'properties' from __all__.
- debug: print_rows/print_columns crashed on query frames (missing add_ranks=False); print_num_rows(by_query=False) passed add_ranks to apply.generic, which rejects it.
- transformer: Indexer.index()/Estimator.fit() were silent no-ops; raise NotImplementedError. 
- apply_base: rename dead transform_output typo to transform_outputs (+docs). 
- Restore dead exception handlers: mavenresolver catches urllib.error.URLError (offline fallback and cache restore now work; local disk errors still propagate); datasets/_core catches requests.exceptions.RequestException.

tests/test_datasets: fetch failures are now wrapped in ValueError, so the network-skip handlers catch it too.


jegentile:
@seanmacavaney, great to learn about this project at SIGIR. Applying some small fixes here and there.

seanmacavaney:
Thanks James! It was great to catch up at SIGIR. Please bear with me as I go through the PR. (Still en route back to UK.)

PR #676, opened 2026-07-27
fix: move to thread paralellism to avoid broken jvm fork state

vlcfaria:
This is a bit of a rough attempt to fix `GridScan` on multiprocessing scenarios.

My understanding of the bug was that `pyterrier_alpha.parallel._parallel_lambda_joblib` uses process-level parallelism, which copies the JVM started flag and the JVM state which can be messy.

I swapped into thread-based parallelism, following [Retriever.transform](https://github.com/terrier-org/pyterrier/blob/9230fe5dc85358027ecda53d330a542e41a270cd/pyterrier/terrier/retriever.py#L465) which I believe shares the JVM more cleanly while maintaining actual parallelism since they release the GIL.  Ive kept the ray backend because it was kind of a grey area.


cmacdonald:
thanks @vlcfaria for this. We dont use the parallel stuff much, which is why it moved to alpha.

Can you explain a bit about your use case?

pyterrier/_evaluation/_grid.py
```diff
@@ -261,9 +259,26 @@ def _evaluate_several_settings(inputs : List[Tuple]):
...
+            except ImportError as ie:
+                raise ImportError("pyterrier-alpha[parallel] must be installed for backend='ray'") from ie
+            eval_list = parallel_lambda(_evaluate_several_settings, batched_inputs, jobs, backend=backend)
+            eval_list = list(itertools.chain(*eval_list))
+        else:
+            # avoid process parallelism by using threads, which share the parent's JVM
+            # instead of calling fork() and leaving leave the JVM in a weird state
+            def _eval_chunk(chunk):
+                out = [_evaluate_one_setting(k, v) for k, v in chunk]
+                detach()  # release JNI refs
```
cmacdonald, line 276: Flake says we should be decorating this method with @pt.java.required, but I think the actual point is that if the transformer doesnt involve Java (Terrier or Anserini) then jnius may not even be installed. So we need to detect jnuis (try import etc) and act accordingly.

pyterrier/_evaluation/_grid.py
```diff
@@ -261,9 +259,26 @@ def _evaluate_several_settings(inputs : List[Tuple]):
...
+            eval_list = parallel_lambda(_evaluate_several_settings, batched_inputs, jobs, backend=backend)
+            eval_list = list(itertools.chain(*eval_list))
+        else:
+            # avoid process parallelism by using threads, which share the parent's JVM
+            # instead of calling fork() and leaving leave the JVM in a weird state
+            def _eval_chunk(chunk):
+                out = [_evaluate_one_setting(k, v) for k, v in chunk]
+                detach()  # release JNI refs 
+                return out
+            with ThreadPoolExecutor(max_workers=jobs) as ex:
```
cmacdonald, line 278: Have you tested that you get a speed up with different Python threads calling Java? I think Terrier's data structures arent thread safe unless they are loaded with "concurrent:" prefix.  At least with the forked JVM, if setup correctly, the results would be correct.

cmacdonald:
> Can you explain a bit about your use case?

Ah, I see the issues now. Usecase is clear. 

I have some questions that I have placed in the review.

